PR #41, opened 2018-08-01
Use has_one :through for Group chair_dept_head

nomadicoder:
- Unable to get administrate to work with has_one / source
- Create a new join table for group contacts


nomadicoder:
Need to customize the Group_dashboard chair_dept_head:HasOne so that it's a pull down rather than an embedded form.

nomadicoder:
@cdoyle-temple Would you take this branch and customize the chair_dept_head field per the previous comment and we'll merge the branch.

cdoyle-temple:
The problem I've run into now is when I try to update a group. I have a select list of people that can be assigned to the chair_dept_head field, but when I submit the form, the select list uses the selected person's ID and the rails app is expecting a person object, not a string ID value

nomadicoder:
@cdoyle-temple I've updated the UI and create/update controllers so we are able to create and modify groups. I'll go head and merge this PR for hands on testing.

PR #51, opened 2018-08-09
Uploads

cdoyle-temple:
Person photo now working.


app/controllers/admin/groups_controller.rb
```diff
@@ -52,5 +52,12 @@ def update
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+
+    def update
```
nomadicoder: Do we need this? If not, we should remove debug override methods.

app/controllers/admin/people_controller.rb
```diff
@@ -17,5 +17,62 @@ class PeopleController < Admin::ApplicationController
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+
```
nomadicoder: Remove the commented out example code.

cdoyle-temple:
Cleaned up the controllers.

nomadicoder:
Thank you, looks good!

PR #52, opened 2018-08-10
Added personal web site to person and updated views.

cdoyle-temple:


nomadicoder:
Should we have a web field that makes the URL clickable in the administrate show page? It might be handy for the administrator so they won't have to copy/paste the URL in another tab to make sure it's a good URL.

PR #56, opened 2018-08-14
Images for models

cdoyle-temple:
Please merge this so I can work on variants in master.


app/fields/building_photo_field.rb
```diff
@@ -0,0 +1,7 @@
+require "administrate/field/base"
+
+class BuildingPhotoField < Administrate::Field::Base
+  def to_s
+	    data
+  end
```
nomadicoder: What will this look like if we do `building.photo.to_s`? Should it be something else, like the name of the file?

app/assets/stylesheets/application.scss
```diff
@@ -104,4 +114,4 @@ form {
 }
 .no-padding {
 	padding: 0;
-}
+}
```
nomadicoder: Please fix - add a new line.

app/fields/person_photo_field.rb
```diff
@@ -0,0 +1,7 @@
+require "administrate/field/base"
+
+class PersonPhotoField < Administrate::Field::Base
+  def to_s
+	    data
+  end
```
nomadicoder: Same comment as the previous `to_s`.

app/fields/space_photo_field.rb
```diff
@@ -0,0 +1,7 @@
+require "administrate/field/base"
+
+class SpacePhotoField < Administrate::Field::Base
+  def to_s
+	    data
+  end
```
nomadicoder: And here too.

app/fields/building_photo_field.rb
cdoyle-temple: building_photo_field has now been deleted. we're not using photos for buildings per cynthia.

app/fields/person_photo_field.rb
cdoyle-temple: Has no visible effect, so am changing as requested.

PR #67, opened 2018-08-16
Collections

cdoyle-temple:
Move group select options into config.


config/application.rb
```diff
@@ -21,5 +21,7 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    config.select_options_for_group = ["Assembly","Committee","Department","Functional Team","Strategic Steering Team","Working Group"]
```
nomadicoder: We should be specific and call it `group_type`.

PR #74, opened 2018-08-23
Refacto to eliminate custom required fields.

bibliotechy:
Required fields can now be defined in the field definintion in a dashboard config.

For example:
```ruby
ATTRIBUTE_TYPES = {
  name: Field::String.with_options(required: true),
  description: DescriptionField.with_options(required: true),
  spaces: Field::HasMany.with_options(required: true),
}

```
This will include a '*' after the label and add the html required and aria-required attributes for that input.

Works for all field types, including custom field types, as long as they inherit from `Administrate::Field::Base`.


nomadicoder:
@bibliotechy - RSpec passed before the merge. After merge, building, person, and monkey_patches specs failed:

```
An error occurred while loading ./spec/dashboards/building_dashboard_spec.rb.
Failure/Error:
  describe BuildingDashboard do
    describe "#tinymce?" do
      it "returns the local override value" do
        expect(BuildingDashboard.new.tinymce?).to be true
      end
    end
  end

NoMethodError:
  undefined method `describe' for main:Object
# ./spec/dashboards/building_dashboard_spec.rb:3:in `<main>'
```

bibliotechy:
Did I need spec helper included?

nomadicoder:
First `describe` blocked needed to be `RSpec.describe`

I pushed up a branch. I'll issue a PR for it.

PR #75, opened 2018-08-23
Fixes broken specs from monkey patch merge

nomadicoder:
- Fixed broken specs that merge induced.  Errors were

An error occurred while loading the committed files. Must use 
`Rspec.describe` block
`

### DIFF
--- a/spec/dashboards/building_dashboard_spec.rb
+++ b/spec/dashboards/building_dashboard_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe BuildingDashboard do #, type: :controller do
+RSpec.describe BuildingDashboard do
   describe "#tinymce?" do
     it "returns the local override value" do
       expect(BuildingDashboard.new.tinymce?).to be true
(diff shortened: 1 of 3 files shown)


spec/dashboards/building_dashboard_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe BuildingDashboard do
+RSpec.describe BuildingDashboard do #, type: :controller do
```
bibliotechy: Do we need the `#, type: :controller do`? Isn't that all commented out?

spec/dashboards/person_dashboard_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe BaseDashboard do
+RSpec.describe BaseDashboard do #, type: :controller do
```
bibliotechy: See above

spec/dashboards/building_dashboard_spec.rb
nomadicoder: I thought I deleted that.

PR #80, opened 2018-08-30
Show a user login page for development

nomadicoder:
- Enable login page for development
- Disable auto-forward for development
- Instructions to add users for development


nomadicoder:
@bibliotechy We need to see if the behavior is the same on Production.

PR #81, opened 2018-08-30
Updates to fix scrolling and page structure issues.

cdoyle-temple:
This should restore the scrolling functionality.


nomadicoder:
Unable to recreate the issue locally - scrolls locally, but not on production. Updated production server, cleared cache, but it still will not scroll.  I'll merge this PR and update production.

PR #82, opened 2018-08-31
Specify OAuth Redirection URI in ENV

nomadicoder:
Currently we are hardcoding the Redirect URI in code that's in source control. This should take care of that problem code by specifying the URI as an environment variable. If not specified, URI defaults to `http://localhost:3000/accounts/auth/google_oauth2`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,15 @@ Fortytude is the Temple University Library's website built on Ruby on Rails.
 ```
 export GOOGLE_OAUTH_CLIENT_ID="Google client ID goes here"
 export GOOGLE_OAUTH_SECRET="Google OAuth secret goes here"
+export GOOGLE_OAUTH_REDIRECT_URI="Google OAuth Redirection URI goes here -- Set below"
 ```
+
 Add these same lines to your `.bash_profile` or `.bashrc` file, depending on
-how you've setup your Bash shell:
+how you've setup your Bash shell.
+
+Set your `GOOGLE_OAUTH_REDIRECT_URI` to the fully qualified domain name of the Rails application server
+Include port, if applicable, and path. If is not set, it will default to
+`http://localhost:3000/accounts/auth/google_oauth2`.
 
 * Clone the repository and navigate to the souce code directory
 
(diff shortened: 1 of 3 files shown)


nomadicoder:
@chad - I have updated the .bashrc on the production server. We need to test this in production.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,5 @@
 module ApplicationHelper
+  def google_oauth_redirect_uri
+    ENV.fetch('GOOGLE_OAUTH_REDIRECT_URI') { 'http://localhost:3000/accounts/auth/google_oauth2' }
```
bibliotechy: Why not just use `"#{root_path}accounts/auth/google_oauth2"`?

nomadicoder: Can you redirect to a file path if you want a URI which is required by Google OAuth?

nomadicoder: We can use `Socket.gethostbyname(Socket.gethostname).first` in production. See: https://stackoverflow.com/questions/151545/how-can-i-get-the-fqdn-of-the-current-host-in-ruby

We can use `localhost:3000` when we're in development.

nomadicoder:
@bibliotechy Get FQDN without using an environment variable.

bibliotechy:
What I really should have pointed you to was using the builtin Rails route helpers [Path and URL Helpers](https://guides.rubyonrails.org/routing.html#path-and-url-helpers)
> Each of these helpers has a corresponding _url helper (such as photos_url) which returns the same path prefixed with the current host, port and path prefix.

So, in this case, we can use `account_google_oauth2_omniauth_authorize_url` directly in the template, with no need for an additional helper.

nomadicoder:
@bibliotechy This is how it's supposed to be done.

README.md
```diff
@@ -13,9 +13,15 @@ Fortytude is the Temple University Library's website built on Ruby on Rails.
 ```
 export GOOGLE_OAUTH_CLIENT_ID="Google client ID goes here"
 export GOOGLE_OAUTH_SECRET="Google OAuth secret goes here"
+export GOOGLE_OAUTH_REDIRECT_URI="Google OAuth Redirection URI goes here -- Set below"
```
bibliotechy: This is no longer need, correct?

bibliotechy: @nomadicoder ping?

nomadicoder:
@bibliotechy We'll get this right some day.  Now using URL helper for the redirect.

nomadicoder:
@bibliotechy README fixed and just has the punctuation fix.

PR #84, opened 2018-09-06
Only allow entering image for admin accounts.

nomadicoder:
Only an admin can upload or modify photos because we have specific constraint requirements on images.


nomadicoder:
@bibliotechy Ready to review passing Travis CI checks.

PR #89, opened 2018-09-11
Create Service class

nomadicoder:
Issue #MAN-21

Creates Service model class.


nomadicoder:
@cdoyle-temple - Would you add admin and views?

nomadicoder:
Assigning User Interface and Administrate portions to @cdoyle-temple

bibliotechy:
Getting an error when trying to save a new Service saying that the group is invalid 
screen shot 2018-09-25 at 8 48 01 pm

even though a group has been selected 
screen shot 2018-09-25 at 8 51 25 pm

bibliotechy:
@nomadicoder Did you see my comment on this?

nomadicoder:
@bibliotechy I ran the spec locally and it passed. I wonder if there's a flaw in the way we're generating related associations. I'm curious if we force a new check to see if it fails. I'd like to see if it's repeatable. Then I can proceed accordingly.

nomadicoder:
@bibliotechy Actually, I've seen this before. It was caused by bad data in the database. I had to drop, remigrate and repopulate my database. I realize this will be a problem if we have production data.

bibliotechy:
Took the steps mentioned and that fixed my issue.

PR #92, opened 2018-09-13
Base `field.required?` on model validator of presence

bibliotechy:
Decide if a field should be required in on the dashboard form if the model (resource) the
dashboard is wrapping contains a valitor of presence for that attribute.

This means we no longer need to explicitly pass the "`required: true` option to field definitions
in dhasboards.


bibliotechy:
@nomadicoder One Build will vener complete here because we cancelled the Travis.com integration.
It's fine to merge without that completing as the others completed successfully.

PR #97, opened 2018-09-26
Create event class

nomadicoder:
Address Issue #MAN-36

Initial go at the event class. Bare, unstyled views. Validations are TBD


spec/views/events/show.html.erb_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
```
bibliotechy: I don't think we need this test at all. We're certainly going to change the display and not include all of the those fields, as some are mutually exclusive.

spec/helpers/events_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
```
bibliotechy: this file does nothing, please delete it.

app/helpers/events_helper.rb
```diff
@@ -0,0 +1,2 @@
+module EventsHelper
```
bibliotechy: Please delete this file as it is not required.

PR #100, opened 2018-09-27
Highlights

cdoyle-temple:
MAN-27

Creates highlight type that can be edited manually in administrate and promoted to homepage which will display top 4 most recent highlights.

Images on homepage will be cropped from center outward to fit 370x196.

Show page currently displays image at full size.


app/controllers/application_controller.rb
```diff
@@ -13,4 +13,8 @@ def get_alert
 	    @alert = Alert.find_by(published: true)
 	end
 
+  def get_highlights
```
bibliotechy: Are highlights going to be on every? Trying to understand when this is in the application controller?

cdoyle-temple: The homepage isn't a model and doesn't have a controller other than application so that's where I put it.

app/views/application/index.html.erb
```diff
@@ -67,59 +66,140 @@
 	</div>
 </div>
 
-<div class="vertical-stripes">
-	<h1> News &amp; Highlights</h1>
+<div class="vertical-stripes" style="margin-top: 63px;text-align: center;width: 100%;">
+	<h1 style="white-space: nowrap;padding:14px;font-size: 250%;letter-spacing:.15rem;background-color:white!important;width: 25%;margin:auto;height:84px"> News &amp; Highlights</h1>
 </div>
 
+<% unless @highlights.empty? %>
```
bibliotechy: All this highlights code feels like it should be broken out into a partial. 

Also, it contains quite a bit of duplication. Do you think it could be refactored to loop over the existing highlights and reuse a single highlight partial?

bin/rails
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env ruby
-begin
```
bibliotechy: what's happening here?

bin/rake
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env ruby
-begin
```
bibliotechy: what's happening here?

cdoyle-temple: I think this can be deleted. there was a problem locally that needed this reset.

bin/rails
cdoyle-temple: I think this can be deleted. there was a problem locally that needed this reset.

PR #107, opened 2018-10-03
Create Event Sync Service

nomadicoder:
- Create Event Sync as a service
- Event model association must be optional (Rails 5 made it required by
  default)
- Elide event fields displayed in default views (to make development life easier)

Rake task to read events from external source

Integrate fuzzy search

Clean up database so we can run different data fixtures without
interference from previous tests


bibliotechy:
Superseded by #114

PR #114, opened 2018-10-15
Create Event Sync Service

bibliotechy:
Create Event Sync as a service

Checks of the event is already present (via hash) or if it needs tobe updated.

Imports images from events site for persistence.

Create rake task to trigger service


bibliotechy:
Supersedes #107

PR #128, opened 2018-10-22
Man 70 ambler pages

cdoyle-temple:
Please see the attached wireframes for the Ambler Campus Library microsite.

One item of discussion that came up in a previous meeting was the idea of using design to differentiate between levels of the site (homepage, entity page, secondary page, etc). I'm not sure exactly what this would look like, but I leave that to you.

(differentiation of secondary homepages accomplished with Big Title on page in green searchbar. Other pages will not have this.)


bibliotechy:
@cdoyle-temple this appears to be built off a a previous pr https://github.com/tulibraries/manifold/pull/127

cdoyle-temple:
It should have the updates from that PR; I (hope I) pulled master after that merge and worked from there.

app/controllers/pages_controller.rb
```diff
@@ -1,10 +1,28 @@
...
   end
 
   def get_highlights
     @highlights = Highlight.where(promoted: true).take(4)
   end
+
+  def ambler
+  end
+
+  def index
```
bibliotechy: I'm not sure why this is here?

app/controllers/pages_controller.rb
```diff
@@ -1,10 +1,28 @@
...
+    @buildings = Building.all
+  end
+
+  def show
```
bibliotechy: I'm not sure why this is here?

app/controllers/pages_controller.rb
```diff
@@ -1,10 +1,28 @@
...
+    @todays_hours = LibraryHours.where(location_id: @building.hours, date: @today)
+  end
+
+  private
+    def set_date
```
bibliotechy: I'm not sure why this is here?

app/controllers/pages_controller.rb
```diff
@@ -1,10 +1,28 @@
 # frozen_string_literal: true
 
 class PagesController < ApplicationController
+  before_action :set_date, only: [:about_ambler]
```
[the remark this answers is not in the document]
cdoyle-temple: Correct, it came and went. Should have cleaned that up.

app/controllers/pages_controller.rb
cdoyle-temple: In my limited experience, I need a def in the controller in order for the route to work.  This may be untrue.

app/controllers/pages_controller.rb
cdoyle-temple: Not sure what you're referring to here.

app/controllers/pages_controller.rb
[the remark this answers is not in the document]
cdoyle-temple: The mini-pages will be unique and a template for them would not be a good fit.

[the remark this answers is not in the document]
cdoyle-temple: Also, not minisites, mini homepage(s). All of the other pages (or most anyway) for a 'minisite' will be generated from existing models, i.e. building page, spaces, services, etc...

app/controllers/pages_controller.rb
cdoyle-temple: pages controller cleaned up.

app/controllers/pages_controller.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 class PagesController < ApplicationController
-  def home
```
bibliotechy: This will break the home page, won't it? The root route is to home

app/controllers/pages_controller.rb
```diff
@@ -1,10 +1,12 @@
...
-  end
 
   def get_highlights
     @highlights = Highlight.where(promoted: true).take(4)
   end
+
+  def index
```
bibliotechy: Your route points to the `:ambler` method, not the index method.

cdoyle-temple:
The thing is, it didn't break anything. I had thought they were necessary
but was confused by your comment asking why they were there. I've out them
back now.

app/controllers/pages_controller.rb
cdoyle-temple: This did not break the homepage.

app/controllers/pages_controller.rb
cdoyle-temple: Not sure where index came from or why it needed all buildings. I may have copied and pasted when setting up the controller. I've now removed the index method as we won't ever need one.

app/controllers/pages_controller.rb
cdoyle-temple: Although I have put it back.

cdoyle-temple:
Re: minisite pages -- they are actually mini-homepages, per se, not minisites. This differentiation has confused several people on our team due to the language used (minisite or microsite, or even minihomepage). Suggest we find another word to label these landing pages. Library pages? -- since they're not building pages and building pages are not necessarily libraries? The homepage for each will be similar to each other, but unique. We won't be able to have a standard template for these pages.

PR #129, opened 2018-10-22
Man 23 updates

cdoyle-temple:
..."That makes sense for building hours. However, for the spaces, if the space itself does not have it's own hours, I would like to pull in the hours from the building."


cdoyle-temple:
Cleaned up the hours methods by moving them to the models and refactoring.

PR #131, opened 2018-10-23
Move services header into conditional check for services.

cdoyle-temple:


cdoyle-temple:
This fails test only on github. It passes locally with no errors. The travis error about the DatabaseCleaner is not in anything I was working on. Neither the master I pulled and is currently up to date, nor the new branch I created (space_fix) has any test errors locally. Please help!

PR #132, opened 2018-10-23
Forms

cdoyle-temple:
These are not finished but cannot be tested until they are on a server with a registered domain name, so that the google gmail servers can be used to send.


cdoyle-temple:
Same error on this branch, DatabaseCleaner. Please help!

PR #139, opened 2018-10-25
Fix Rubocop Issues

nomadicoder:
- Fixes style issues that were in Master branch.
- Updates schema date


nomadicoder:
@bibliotechy - Would you review and merge so we won't have to make the same changes on each of our branches.

PR #146, opened 2018-10-30
Add model, controller & migration for BlogPosts

bibliotechy:
Connects blog posts to individual Blog

Updates the Person fixture to handle its own dependent associations.

### DIFF
--- a/spec/factories/persons.rb
+++ b/spec/factories/persons.rb
@@ -13,8 +13,7 @@ FactoryBot.define do
     springshare_id { "0123-4567-8901" }
     # Add related objects in create.
     # e.g.
-    #   let(:building) { FactoryBot.create(:building) }
-    #   let(:space) { FactoryBot.create(:space, building: building) }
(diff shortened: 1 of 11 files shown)


spec/factories/persons.rb
```diff
@@ -13,8 +13,7 @@
     springshare_id { "0123-4567-8901" }
     # Add related objects in create.
     # e.g.
-    #   let(:building) { FactoryBot.create(:building) }
-    #   let(:space) { FactoryBot.create(:space, building: building) }
+    spaces { [FactoryBot.create(:space, building: FactoryBot.create(:building))] }
```
bibliotechy: @skng5 Check this out. Was able to set up the person factory to handle creating it's own dependencies by inlining the calls to factorybot. Ugly, but less ugly than having them in every singe place where we use the person factory.

nomadicoder: Nice! Thank you.

PR #148, opened 2018-10-31
Maybe this extra cleaner is what is causing the issue?

bibliotechy:


nomadicoder:
@bibliotechy thanks. I wonder why I did that.

PR #150, opened 2018-10-31
Man 84 groups show final

cdoyle-temple:
Please apply design and code to the group entity pages.

10/25 (gcd) Change dept_chair_head field to has_many. (handled through MAN-89 sub task)


cdoyle-temple:
I get no errors with this locally. Can someone help me out?

PR #158, opened 2018-11-05
Groups: removed lingering phone and email from schema

cdoyle-temple:
The phone and email fields for groups should have also been removed from the schema.


bibliotechy:
We're going to leave these in as we are not actually removing that from Schema at the moment, just not using them.

PR #176, opened 2018-11-16
Man 69 design mock up for library hours

cdoyle-temple:
#MAN-69

Please create a design mockup for the library hours. Attached is a wireframe for 1) the landing page of all hours and 2) the hours for a specific library/space. 20181003-Hours.pdf


app/controllers/library_hours_controller.rb
```diff
@@ -1,24 +1,108 @@
...
     @today = Date.today
-    sunday = @today.beginning_of_week - 1
-    saturday = @today.end_of_week
+    @cyear = @today.year
+    @cweek = @today.cweek
+
+    @week =  params[:week].nil? ? @cweek : params[:week].to_i
+    @year = params[:year].nil? ? @cyear : params[:year].to_i
+
+    @first_week = LibraryHours.where(location_id: "paley").first.date.to_date.cweek
```
bibliotechy: We're doing four database queries here on every page, even if we don't need to know about next week. It feels like we should only calculate that on demand, not every time

app/controllers/library_hours_controller.rb
```diff
@@ -1,24 +1,108 @@
...
+    @first_year = LibraryHours.where(location_id: "paley").first.date.to_date.year
+    @last_week = LibraryHours.where(location_id: "paley").last.date.to_date.cweek
+    @last_year = LibraryHours.where(location_id: "paley").last.date.to_date.year
+
+    @next_week = @week + 1
+    @prev_week = @week - 1
+
+    if @week == 52
```
bibliotechy: it feels like a lot of this could be handled by rails built in date extensions, like `.next_week`

```
t = Date.parse("31/12/2018")
=> Mon, 31 Dec 2018
t.next_week
=> Mon, 07 Jan 2019
```

PR #202, opened 2018-12-05
MAN-27 News and highlights section

cdoyle-temple:
#MAN-27

Create four news and highlights slots where Sara and Kaitlyn can manually add news and highlight items. There should be a link, an image and text. This is similar to the manual spots you created on the TUpress site.

These will likely need to be stored locally on the server, as opposed to being streamed in from WordPress, LibGuides, etc.

*****************

- Updated fields for highlights and adjusted show page to display on site rather redirect to external or other internal url.

- Cleaned up tests and rubocop

### DIFF
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -86,7 +86,7 @@
 				  <div class="card-body">
 				    <span class="card-title"><%= highlight.title %></span>
 				    <p class="card-text"><%= highlight.blurb %></p>
-				    <%= link_to "See more >", highlight.link, class: 'see-more' %>
+				    <%= link_to "See more >", highlight_path(highlight), class: 'see-more' %>
 				  </div>
 				</div>
 			</div>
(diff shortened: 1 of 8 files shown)


app/views/pages/home.html.erb
```diff
@@ -86,7 +86,7 @@
 				  <div class="card-body">
 				    <span class="card-title"><%= highlight.title %></span>
 				    <p class="card-text"><%= highlight.blurb %></p>
-				    <%= link_to "See more >", highlight.link, class: 'see-more' %>
+				    <%= link_to "See more >", highlight_path(highlight), class: 'see-more' %>
```
bibliotechy: Just making sure this requirement changed. We now want to go to the show page for each highlight, and not to the link that was included when the highlight was created?

cdoyle-temple: Cynthia was insistent about this.

PR #220, opened 2018-12-14
Removed fill an associated bubbles and lessened roundness.

cdoyle-temple:


bibliotechy:
@cdoyle-temple can you resolve these conflicts?

PR #241, opened 2019-01-02
Homepage Responsive Design

cdoyle-temple:
Using the mock-up for the homepage, please create a mobile version. You can use an iPhone sample device.

***** NOTE: this adversely affects the mini-homepages responsive design, but since their entire design is being redone, this should be ok and future tickets will address their responsiveness ******

- Add responsive styling for Highlights section of homepage
- Add responsive styling for static  homepage blocks
- Update search options and hours info for responsive design
- Updates to pass tests (add strip to description cleaner) and fix some alignments.
- updates to hsl and ambler homepages to not render quite so badly at low res

### DIFF
--- a/app/views/application/_search-options.html.erb
+++ b/app/views/application/_search-options.html.erb
@@ -1,0 +1,23 @@
+<div id="searchOptions">
+  <ul>
+    <li><%= link_to "Everything", "#" %></li>
+    <li><%= link_to "Books", "#" %></li>
+    <li><%= link_to "Articles", "#" %></li>
+    <li><%= link_to "Databases", "#" %></li>
+  </ul>
+  <div id="advancedSearchLink">
+    <%= link_to "Advanced Search", "#" %>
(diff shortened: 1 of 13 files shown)


app/views/application/_search-options.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <li><%= link_to "Databases", "#" %></li>
+  </ul>
+  <div id="advancedSearchLink">
+    <%= link_to "Advanced Search", "#" %>
+  </div>
+</div>
+
+<div id="dateTime">
+  <div id="date">
+    FRIDAY, AUGUST 17, 2018
```
bibliotechy: @cdoyle-temple JUst double checking that this is supposed to be hard coded for now?

bibliotechy: @cdoyle-temple nevermind. I see where it is pulled from

PR #253, opened 2019-01-09
#MAN-28 Entities Add To Footer Link

cdoyle-temple:
#MAN-28

Add a field to the entity model for buildings, services, groups and collections labeled "Add to footer".

Field should be a check box.

Field is optional.

Field is an admin-only change.

The "Add to footer" checkbox will designate items that appear in the footer.

************************************

- Added add_to_footer checkbox to four footer models
- updated application controller and footer partial to only pull checked instances.
- rubocop, test fixes


bibliotechy:
Registering my hesitation about adding four SQL queries to the Footer, but...not going to block it because of that.

PR #257, opened 2019-01-10
MAN-145 Add dashboard for blogs

nomadicoder:
* NOT READY TO MERGE *


app/services/sync_service/blogs.rb
```diff
@@ -0,0 +1,75 @@
...
+    blog_post_hash.merge!(author(blog_post))
+  end
+
+  def create_or_update_if_needed!(record_hash)
+    # If a record already exists with this content hash, then no update needed
+    unless BlogPost.exists?(content_hash: record_hash["content_hash"])
+      # Fuzzy find record based on title and start time, and otherwise create a new one
+      blog_post = FuzzyFind::BlogPost.find(
+        record_hash["title"],
+        addl_attribute: { post_guid: record_hash["post_guid"] }
```
bibliotechy: So...do we need to use fuzzy find here? The post guid isn't a possibly variable string like "Steven Ng" or "Steve Ng", it is a foreign key, so we should probably use it as such. I think the logic here is something like

```
blog_post = BlogPost.find_by(blog: @blog_id, post_guid: record_hash["post_guid"]) || BlogPost.new
blog_post.assign_attributes(record_hash)
```

app/dashboards/blog_post_dashboard.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
```
bibliotechy: I'm not sure there is a requirement for this Dahsboard. I don't think we'll be doing any editing of Blog posts here - it all needs to happen upstream in Wordpress.

spec/services/fuzzy_find/finder_service_spec.rb
```diff
@@ -111,7 +110,7 @@ def self.where(args)
                   needle: "Newish Person",
                   haystack_model: Person,
                   attribute: :name,
-                  addl_attribute: { groups: "1" }
+                  addl_attribute: { groups: ["1"] }
```
bibliotechy: So, I think changing this is causing tests to fail on Travis. What's going on here? Why did it change?

app/services/sync_service/blogs.rb
nomadicoder: Replaced `FuzzyFind::BlogPost` with `BlogPost.find_by`.

app/dashboards/blog_post_dashboard.rb
nomadicoder: Removed.

app/services/sync_service/blogs.rb
```diff
@@ -0,0 +1,72 @@
...
+      "blog" => Blog.find(@blog_id),
+      "content_hash" => xml_hash,
+    }
+    blog_post_hash.merge!(author(blog_post))
+  end
+
+  def create_or_update_if_needed!(record_hash)
+    # If a record already exists with this content hash, then no update needed
+    unless BlogPost.exists?(content_hash: record_hash["content_hash"])
+      # Fuzzy find record based on title and start time, and otherwise create a new one
```
bibliotechy: cop
```suggestion
```

PR #259, opened 2019-01-15
#MAN-107 Added ability to display units for LRS department.

cdoyle-temple:
#MAN-107

Create a new field for "home department". This field will be a related group and will be the person's home department. The "Groups" field will be used to list any other groups that this person belongs to that are secondary to their home department.

The contact information that displays on the person's page will be pulled from the 'Home department' related space information.

******************************

Added ability to display units for LRS department. Updated person index and show view templates to pull in the units and display if the dept is LRS.

### DIFF
--- a/app/views/persons/index.html.erb
+++ b/app/views/persons/index.html.erb
@@ -47,7 +47,14 @@
     <div class="col-3">
       <% person.groups.each_with_index do |group, index| %>
         <% if group.group_type == "Department" %>
-          <%= group.name %><% unless index == (person.groups.size - 1) %>, <br /><% end %>
+          <%= group.name %><% if group.name == "Learning and Research Services" %>: 
+            <% person.groups.each do |unit| %>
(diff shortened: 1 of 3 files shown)


app/views/persons/index.html.erb
```diff
@@ -47,7 +47,14 @@
     <div class="col-3">
       <% person.groups.each_with_index do |group, index| %>
         <% if group.group_type == "Department" %>
-          <%= group.name %><% unless index == (person.groups.size - 1) %>, <br /><% end %>
+          <%= group.name %><% if group.name == "Learning and Research Services" %>:
```
bibliotechy: Hard coding the name of a unit here is probably not  a great idea. Seems like we are missing something in the data model to handle a Parent Group relationship.

cdoyle-temple: This was supposed to be a one-off and wouldn't be repeated, but overheard something this morning that leads me to believe otherwise, on other models (namely Services).

PR #260, opened 2019-01-16
MAN-159 add new event tests

nomadicoder:
- Tests added to validate sync_event methods.
- Event feed has events with start and end times indicating "(All day)". ISO-8601 has now way to indicate that an event has not time, or is for a full day.  Modification adds boolean all_day tag field. If tag is set, then the time for the event is rendered "(All day)" otherwise it renders the given start and end time.


nomadicoder:
@bibliotechy Ready for review.

PR #263, opened 2019-01-17
MAN-128 Create Exhibition enitity model

cdoyle-temple:
*NOTE:* I'm not sure why the Events works @skng5 did is included in this commit, but it shouldn't overwrite anything he did as I didn't change any of it.

Create a model with the following fields:

Title - single line text, required, non-repeatable
Description - paragraph text, required, non-repeatable
Image - file upload, optional, non-repeatable
Begin date - date field, optional, non-repeatable
End date - date field, optional, non-repeatable

Related spaces - one exhibit can have only one space
Related groups - one exhibit can have many groups
Related collection - one exhibit can have many collections

************************

- Created model with specified attributes
- Updated show view
- Updated tests


config/routes.rb
```diff
@@ -2,13 +2,15 @@
 
 Rails.application.routes.draw do
 
+  resources :exhibitions
```
bibliotechy: I think this one is extra

spec/routing/exhibitions_routing_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
```
bibliotechy: These looks like Scaffold specs that actually shouldn't be passing. We don't want update via PUT to work for example. But having the full resources route noted above would make them pass. 

The only routes that should be available are index and show, as everything else is handled by administrate.

config/routes.rb
cdoyle-temple: removed.

PR #274, opened 2019-01-22
MAN-54 Add sync selected blog capability to dashboard

nomadicoder:
Addresses Issue #MAN-54 to add ability to sync a single blog selected from the dashboard.


bibliotechy:
Looks good. Only feedback I have is that :last_sync_date doesn't need to be a editable field in the form. That should never be set by the user.

PR #278, opened 2019-01-25
#MAN-202 Remove space between entities

cdoyle-temple:
#MAN-202 
Similar to what we did on the main homepage, can you remove the space between the building entities on the index page and instead have more white space on the outside of the three columns.

*********************
- Changed rows of three to rows of four in order to minimize space betw…een images when window stretches.
- Also optimized calls to and display of images for index and show pages.

### DIFF
--- a/app/assets/stylesheets/buildings.scss
+++ b/app/assets/stylesheets/buildings.scss
@@ -22,7 +22,8 @@ div.required  label::after {
 		margin: auto;
 		background-color: $background-grey;
 		max-width:680px;
-		height:450px;
+		height: auto;
+		min-height:450px;
 		div.text {
 			padding:21px;
 			margin:auto;
@@ -77,7 +78,14 @@ div.required  label::after {
     overflow: hidden;
     padding: 0;
     width: 250px;
-    height: 190px;
+    height: 189px;
+    .lib-image {
+    	/*min-height:190px;*/
+    }
+    .default-image {
+    	width:190px;
+    	height:190px;
+    }
 	}
   .card-title {
   	font-size:125%;
@@ -95,11 +103,11 @@ div.required  label::after {
 		position: relative;
 		line-height: 3.0rem;
   	display: table;
-			a {
-				color: $white;
-			};
 		font-size: larger;
 		margin-bottom: 21px;
+		a {
+			color: $white;
+		}
 	}
 	.card-body {
     display: table-cell;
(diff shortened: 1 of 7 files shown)


app/models/building.rb
```diff
@@ -20,6 +21,17 @@ class Building < ApplicationRecord
   before_validation :normalize_phone_number
   before_validation :sanitize_description
 
+  def index_image
+    variation =
+      ActiveStorage::Variation.new(Uploads.resize_to_fill(width: 250, height: 190, blob: photo.blob, gravity: "Center"))
```
nomadicoder: Should we parameterize the width and height somewhere?

app/models/building.rb
```diff
@@ -20,6 +21,17 @@ class Building < ApplicationRecord
...
+    ActiveStorage::Variant.new(photo.blob, variation)
+  end
+  def show_image
+    variation =
+      ActiveStorage::Variation.new(Uploads.resize_to_fill(width: 600, height: 450, blob: photo.blob, gravity: "Center"))
```
nomadicoder: Parameterize literal values?

app/assets/stylesheets/buildings.scss
```diff
@@ -22,7 +22,8 @@ div.required  label::after {
 		margin: auto;
 		background-color: $background-grey;
 		max-width:680px;
-		height:450px;
+		height: auto;
+		min-height:450px;
```
nomadicoder: Parameterize 450px somewhere?

app/assets/stylesheets/buildings.scss
```diff
@@ -77,7 +78,14 @@ div.required  label::after {
     overflow: hidden;
     padding: 0;
     width: 250px;
-    height: 190px;
+    height: 189px;
```
nomadicoder: Parameterize 189px somewhere?

app/models/building.rb
cdoyle-temple: It's currently this way all over the site. That would be a bigger ticket than this. Can we do it there (in the new ticket) if necessary?

app/assets/stylesheets/buildings.scss
cdoyle-temple: This needs to be this way in the css.

app/assets/stylesheets/buildings.scss
cdoyle-temple: This needs to be this way in the css.

app/models/building.rb
cdoyle-temple: See above comment.

app/models/building.rb
nomadicoder: Thanks. I'll resolve this.

PR #279, opened 2019-01-25
#MAN-107 Create a "home department" group in Person Entity

cdoyle-temple:
#MAN-107 Create a "home department" group in Person Entity

Create a new field for "home department". This field will be a related group and will be the person's home department. The "Groups" field will be used to list any other groups that this person belongs to that are secondary to their home department.

The contact information that displays on the person's page will be pulled from the 'Home department' related space information.

Added the ability to add another group to a particular group, which would then be the parent group and the cild groups can be parsed out in the display(s).


bibliotechy:
@cdoyle-temple OK, I added the parent group, and conversely, child_groups. Should be saving now on the administrate form. Does not limit scope to prevent a group from being it's own parent.

PR #283, opened 2019-01-30
#MAN-122 -- Finding Aid: "model only"

cdoyle-temple:
#MAN-122 Create Finding Aid Model

Create a model for "Finding Aid" with the following fields:

*Fields: *
name - single line text, required, not-repeatable
description - paragraph text, required, not-repeatable
subject - select from list - see sub-task for list, optional, repeatable
content_link - single line text (primarily used for links), optional, not-repeatable
identifier - single line text, optional, not-repeatable

Related Entities
collection - one resource belongs to one collection, 
person - one resource can belong to many people, but will usually belong to just one person.

**********************************

Create finding aid model, only show view, with associations for an owning collection and a responsible person (multiple).


app/controllers/services_controller.rb
```diff
@@ -13,7 +13,11 @@ def index
   # GET /services/1
   # GET /services/1.json
   def show
-    @related_services = Service.where(service_category: @service.service_category).where.not(id: @service.id)
```
bibliotechy: These changes seem unrelated to Finding Aid?

app/dashboards/person_dashboard.rb
```diff
@@ -20,7 +20,7 @@ class PersonDashboard < BaseDashboard
     chat_handle: Field::String,
     photo: PhotoField.with_options(admin_only: true),
     job_title: Field::String,
-    specialties: Field::MultiSelectField.with_options(
```
bibliotechy: These changes seem unrelated to Finding Aid?

app/views/services/show.html.erb
```diff
@@ -7,17 +7,19 @@
 </div>
 
 <div class="row ctaRow service-show">
-  <% unless @related_services.blank? %>
+  <% unless @services.blank? %>
```
bibliotechy: These changes seem unrelated to Finding Aid?

db/schema.rb
```diff
@@ -164,6 +164,32 @@
     t.index ["space_id"], name: "index_exhibitions_on_space_id"
   end
 
+  create_table "finding_aid_responsibilities", force: :cascade do |t|
```
bibliotechy: Does not match the migration

app/controllers/services_controller.rb
cdoyle-temple: Reverted this. I think these come from other work. Not sure why included here.

app/dashboards/person_dashboard.rb
cdoyle-temple: Without this namespace, I was getting administrate errors due to association of finding aid with person model. Setting the namespace resolved the error.

app/controllers/services_controller.rb
cdoyle-temple: Looks like the line was just moved.

app/dashboards/person_dashboard.rb
cdoyle-temple: Person still works as expected.

app/controllers/services_controller.rb
cdoyle-temple: This and the one below for services show page are work that was aborted and it seems the branch was not reset properly before continuing work on unrelated issue of finding aids. Sloppy work. Sorry.

db/schema.rb
cdoyle-temple: Removed errant table.

app/views/services/show.html.erb
cdoyle-temple: See comment above re: branch cleanup

db/migrate/20190129185647_create_finding_aid_model.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.string :name
+      t.text :description
+      t.string :subject
+      t.string :content_link
+      t.string :identifier
+      t.references :collection, foreign_key: true
+
+      t.timestamps
+    end
+    create_table :finding_aid_responsibilities do |t|
```
bibliotechy: This is plural `finding_aid_responsibilities` while the schema still has the singular `finding_aid_responsibility`

app/controllers/services_controller.rb
cdoyle-temple: Updated to current master version.

db/migrate/20190129185647_create_finding_aid_model.rb
cdoyle-temple: The schema was wrong, it should be pluralized.

PR #302, opened 2019-02-11
#MAN-208 Code event/exhibit index page

cdoyle-temple:
#MAN-208

This design was reviewed with communications staff and agreed upon

***************************************

Add Kaminari for pagination (chose this gem for compatibility with Administrate)
Add event_type field to model since tags is not where Event Type comes from in the xml feed
Update event sync service to pull type field and map to event_type
Add promoted to events checkbox to exhibition model to pull only those approved to the events page
Serialized tags and types with methods in model
Added methods to model to call image variants from view
Updated routes to remove beyondthepage reference and revert

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -39,15 +39,14 @@ class Event < ApplicationRecord
   end
 
   def index_image
-    image.variant(centered_image_variation).processed
-  end
-  
-  def show_image
-    image.variant(centered_image_variation).processed
+    image.variant(centered_image_variation(220, 220)).processed
   end
 
-  def centered_image_variation
-    ActiveStorage::Variation.new(Uploads.resize_to_fill(width: 220, height: 220, blob: image.blob, gravity: "Center"))
-    
+  def show_image
+    image.variant(centered_image_variation(300, 300)).processed
+  end
+
+  def centered_image_variation(width, height)
+    ActiveStorage::Variation.new(Uploads.resize_to_fill(width: width, height: height, blob: image.blob, gravity: "Center"))
   end
 end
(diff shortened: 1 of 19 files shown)


cdoyle-temple:
This should be good now, as long as using Select is ok. 
We will have data issues due to the fact that the feed uses Paley Library and we don't have Paley in the new site. Spaces with buildings not in the site will appear as external spaces.

db/schema.rb
```diff
@@ -171,8 +173,8 @@
     t.integer "person_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["finding_aid_id"], name: "index_finding_aid_responsibilities_on_finding_aid_id"
-    t.index ["person_id"], name: "index_finding_aid_responsibilities_on_person_id"
+    t.index ["finding_aid_id"], name: "index_finding_aid_responsibility_on_finding_aid_id"
```
bibliotechy: Why did this change?

cdoyle-temple: It does not look there has been a change. The lines removed are added back just as they were, and in the same place. I.e. I don't know, but it shouldn't have any effect.

app/models/event.rb
```diff
@@ -27,4 +37,15 @@ def set_times
       "All Day"
     end
   end
+
+  def index_image
+    variation =
+      ActiveStorage::Variation.new(Uploads.resize_to_fill(width: 220, height: 220, blob: image.blob, gravity: "Center"))
```
bibliotechy: @cdoyle-temple I need you to walk me through what is happening here. If I understand this correctly, you are generating the image to that specified size each time. Whch means if there are 10 images on a page, we are generating them 10 times.

cdoyle-temple: yes. is there a better way?

PR #310, opened 2019-02-12
#MAN-221 Charles Library Hours Spreadsheet

cdoyle-temple:
#MAN-221

Steven has requested to have the following column added to the New Charles
Library spreadsheet: Service Zone

NOTE:  Can you please make all  the columns the same width because the SCRC
column (plus café looks like it might be too short too) is cutting off
information.  The spreadsheet won’t allow me to make the change(s) and says
to contact spreadsheet owner.

*********************************
Resized columns
Added new column
Update hours app to use new Charles based location names and hours

### DIFF
--- a/app/controllers/library_hours_controller.rb
+++ b/app/controllers/library_hours_controller.rb
@@ -1,22 +1,21 @@
 # frozen_string_literal: true
 
 class LibraryHoursController < ApplicationController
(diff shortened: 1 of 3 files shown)


bibliotechy:
So, this looks fine, but I thought only Sheryl edits this Spreadsheet. Steven edits it too?

Ugh...I hate this so much

PR #312, opened 2019-02-12
#MAN-220 Update routes for Collections, Services, and Policies

cdoyle-temple:
#MAN-220 Update routes for Collections, Services, and Policies

Remove the index route from Collections, Services, and Policies as the index will be a sidebar on the show page.

***********************************

remove index routes for services, collections, and policies

### DIFF
--- a/app/views/application/_footer.html.erb
+++ b/app/views/application/_footer.html.erb
@@ -36,7 +36,7 @@
             </ul>
         </div>
         <div class="col-md-6 col-lg-3 footer-col" id="footer-services">
-          <h2><%= link_to "Services", service_path(2), class: "col-heading" %></h2>
+          <h2><%= link_to "Services", services_path, class: "col-heading" %></h2>
           <ul>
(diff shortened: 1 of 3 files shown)


bibliotechy:
Not going to merge any PRs that hard code IDs.

bibliotechy:
Didn't intend to close this. Just meant to comment.

cdoyle-temple:
Removed those from this PR. They are still in PR #303 and we can solve the issue there. They don't need to be in this to accomplish what is needed for the issue.

cdoyle-temple:
Removing the hard coding of IDs leaves the entity path (services_path) as undefined since we removed the index page for Services and put the list of all Services on the Show page for each Service, hoping to use one specific Service as the entry point if not ID is passed. Putting the index route back for these will resolve this error but re-introduce another: /services will now present an error to the user since there no longer is an Index view. 

Should we meet with Cynthia since she has spec'd out direct links to the index page for these entities in the site footer?

PR #314, opened 2019-02-13
Keep version below 5.0.0 (4.3.0).

cdoyle-temple:


Gemfile
```diff
@@ -21,7 +21,7 @@ gem "administrate"
 #gem "tustyles", git: "https://github.com/tulibraries/tustyles"
 gem "bootstrap", "~> 4.2.1"
 gem "popper_js", "~> 1.14.5"
-gem "tinymce-rails"
+gem "tinymce-rails", "4.3.0"
```
bibliotechy: I'd actually prefer this was `~> 4.3` so we keep getting minor updates without getting major updates.

```suggestion
gem "tinymce-rails", "~> 4.3"
```

cdoyle-temple: Sure, did not know minor updates still happened to a prior major version.

PR #317, opened 2019-02-15
MAN-213 track versions of entities

nomadicoder:


nomadicoder:
Needs test for Whodunit

nomadicoder:
@bibliotechy @cdoyle-temple Ready for review.

app/controllers/application_controller.rb
```diff
@@ -5,11 +5,7 @@ class ApplicationController < ActionController::Base
   # user, we have no access to the devise object. So, we need to override
   # current_user to return the current account. This is needed both
   # in ApplicationController and in Admin::ApplicationController
-  before_action :get_alert, :set_footer
-
-  def current_user
-    current_account
-  end
+  before_action :get_alert, :set_footer, :set_paper_trail_whodunnit
```
bibliotechy: Aren't get alert and set footer just used for display? Do they need whodunnits?

db/schema.rb
```diff
@@ -199,8 +199,6 @@
   create_table "groups", force: :cascade do |t|
     t.string "name"
     t.text "description"
-    t.string "phone_number"
```
bibliotechy: @skng5  these keep coming up in your PRs. Why?

app/controllers/application_controller.rb
nomadicoder: @bibliotechy According to the documentation, this is where it's supposed to go

db/schema.rb
nomadicoder: @bibliotechy I may have to start with a fresh working repo.

nomadicoder:
@bibliotechy please review.

PR #318, opened 2019-02-15
#MAN-211 Create Page Entity

cdoyle-temple:
#MAN-211 Create Page Entity

I want a model called "pages" to use for things like the Dean's Welcome, the Diversity and Inclusion statement, and other various administrative statements that are different from "Policies". I envision the Page to also be used for various document uploads, for example: the org chart documents, the annual report, past version of the Speaking Volumes newsletter, etc.

Fields:

Title - single line text, non-repeatable, required
Description/statement text - paragraph WYSIWYG editor, non-repeatable, required
Document upload - restrict fil type to PDF, optional, non-repeatable

Related Entities:
Group - one statement could have multiple groups, one group can have multiple statements, optional

**************************************

- Create page model with associated groups and file attachments;
- Create dashboard
- Add layouts chosen per page from preset list of layouts (like themes in other CMS's).


app/models/page.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Page < ApplicationRecord
+  include Validators
+  has_many_attached :documents, dependent: :destroy
```
bibliotechy: I thought we decided to do this as `has_one`? We'd have a page object for each org chart, rather than a page with many org charts.

db/schema.rb
```diff
@@ -146,6 +146,7 @@
     t.string "ensemble_identifier"
     t.text "tags"
     t.boolean "all_day", default: false
+    t.string "event_type"
```
bibliotechy: What is this?

db/migrate/20190215145600_create_pages.rb
```diff
@@ -0,0 +1,16 @@
...
+class CreatePages < ActiveRecord::Migration[5.2]
+  def change
+    create_table :pages do |t|
+      t.string :title
+      t.text :description
+      t.string :layout
+      t.references :group, foreign_key: true
+      t.timestamps
+    end
+    change_table :groups do |t|
```
bibliotechy: This would mean that a group can only have one page, I don't think that's the expectation. 

We need a join table.

cdoyle-temple: Groups should be able to have multiple pages.

app/models/page.rb
cdoyle-temple: There may be instances of a page that would need more than one file attachment.

db/schema.rb
cdoyle-temple: We should have types of pages so that we can tell how to render, if adding attachment link on the page, or just displaying the attachment (as in orgchart displays). I've added associated template types in application.rb but they don't seem pulled into this pr. Will add them. This gives the ability to have different types of pages with different layouts, as needed. I thought we had talked about this in the meeting with Cynthia.

app/models/page.rb
cdoyle-temple: Orgcharts come to mind, where we can list all the charts on one page, like an index, and then link to them.

db/migrate/20190215145600_create_pages.rb
cdoyle-temple: So yeah, what I have is  just the ability to add multiple pages to a group, and multiple groups to a page, but the association is not there.

db/schema.rb
bibliotechy: OK, this makes sense. 👍

db/migrate/20190215145600_create_pages.rb
cdoyle-temple: Updated page to have belongs_to group, and group still has_many pages.

cdoyle-temple: Fixed associations using new join table group_page. Has many to has many on groups and pages.

PR #325, opened 2019-02-22
Additional work requested from testing. #MAN-154

cdoyle-temple:
Refactor index sidebar to actually work. 
Fix icons.


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_02_08_195724) do
+ActiveRecord::Schema.define(version: 2019_02_15_183324) do
```
bibliotechy: @cdoyle-temple I don't think any of these schema chnages need to be part of this PR, do they?

cdoyle-temple: no.

cdoyle-temple: I think they are there because I built on previous work thinking it would be merged first. They don't need to be here.

PR #338, opened 2019-03-13
#Man-232 export finding aids from drupal

cdoyle-temple:
#Man-232 export finding aids from drupal

Please export the current finding aids from Drupal in a format that can be ingested into Rails. If the export format cannot be automatically ingested, then transformation work may be required.

1. Download json files for import to root directory from: 
- library.temple.edu/scrc/collections/pjac/export/json
- library.temple.edu/scrc/collections/allbutpjac/export/json
(note: reason for two files is that all finding aids as one file caused errors)
2. Run rails db:populate:collections (adds test collections required by finding aids ingest)
3. Run rails db:populate:finding_aids

********************************************* 

- Created task to ingest json data from drupal
- Created task to create collections for working locally (db:populate:finding_aids)
- Added drupal_id field for possible troubleshooting after ingest
- Removed redundant multiple true, and include_blank in dashboards (handled on the custom field level for all)
- Updated show view to handle array rather than string for subjects
- Updatedshow view for possible xss attack
- Updated model to remove nil values coming through on form submission as Subject values.


cdoyle-temple:
I have inadvertantly pushed another branch into this branch. I will need to make sure this branch is back to its latest state.

lib/tasks/hours_ingest.rake
```diff
@@ -1,65 +0,0 @@
-# frozen_string_literal: true
```
bibliotechy: @cdoyle-temple Why is this file deleted in this PR?

cdoyle-temple: The file was removed from this branch as it should not have been there.

cdoyle-temple: Hours ingest is another pending PR.

cdoyle-temple: I had inadvertently pushed the hours refactor branch into this one and the removal of this file is part of the cleanup of that.

PR #339, opened 2019-03-13
Use env var for FQDN in production

bibliotechy:
Rails needs to know the fqdn to set a root_url, so we need to set that in a config via an env variable.


cmharlow:
Any reason this is sitting? Who should review? @bibliotechy

PR #347, opened 2019-03-20
#MAN-55 Refactor Hours implementation

cdoyle-temple:
#MAN-55 Refactor Hours implementation

- move the logic into a service Object
- Call the service object from Rake
- Allow the service object to be called from a Admin Dashboard.
- Add tests for the service object

Update hours controller and views
Move google logic to service
Add sync button to new hours dashboard
Correct singularization of LibraryHours to Library Hour (helpers and tests)

Everything except the test on the service

### DIFF
--- a/app/controllers/admin/library_hours_controller.rb
+++ b/app/controllers/admin/library_hours_controller.rb
@@ -17,5 +17,12 @@ module Admin
     def valid_action?(name, resource = resource_class)
       %w[new edit destroy].exclude?(name.to_s) && super
     end
+
+    def index
+      # get the most recently updated record to find out when
+      # the last sync happened
+      @last_sync = LibraryHour.order(:updated_at).take
+      super
+    end
   end
 end
--- a/app/controllers/admin/library_hours_controller.rb
+++ b/app/controllers/admin/library_hours_controller.rb
@@ -3,9 +3,14 @@
 module Admin
   class LibraryHoursController < Admin::ApplicationController
     def sync
-      SyncService::LibraryHours.call
-      flash[:notice] = "Hours synced"
-      redirect_to admin_library_hours_path
+      begin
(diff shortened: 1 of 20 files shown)


app/controllers/admin/library_hours_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Admin
+  class LibraryHoursController < Admin::ApplicationController
+    def sync
+      SyncService::LibraryHours.call
+      flash[:notice] = "Hours synced"
```
bibliotechy: You need to handle the error case where sync service didn't work.

app/dashboards/library_hour_dashboard.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
```
bibliotechy: I'm not sure we want this dashboard at all. No one should be editing hours here. I know it exists just to have a sync button. Maybe what we need is a completely overridden Dashboard page that just has a big sync button and lists the last time a sync was run (based on the date with the most recent update_date time?

bibliotechy: Not going to block this PR for that, but going to create a new issue based off this idea.

spec/models/library_hour_spec.rb
```diff
@@ -32,7 +32,7 @@
         library_hour = FactoryBot.create(:library_hour, k => v.first)
         library_hour.update(k => v.last)
         library_hour.save!
-        expect(library_hour.versions.last.changeset[k]).to match_array(v)
+        # expect(library_hour.versions.last.changeset[k]).to match_array(v)
```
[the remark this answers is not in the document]
cdoyle-temple: We're not versioning hours. It's not needed and should be removed.

PR #368, opened 2019-04-05
#MAN-208 code event exhibit index page

cdoyle-temple:
#MAN-208 

- Update events_controller.rb
- refactor controller and views, update routes, to better accommodate past events
- Add condition to check for All Day string which is not accepted as datetime db field
- Update times logic for views to method on model
- Update Location facets to link to entities and external locations
- Update Types facets to always display all types, same as Location facets

Load past events from Drupal generated xml at: library.temple.edu/events/xml, must be downloaded and first blank lines removed so that the xml declaration line is first in the file, and find and replace all &quot; with " or image routine will fail, we only need this done once per environment (dev, prod, local)

This is due to Drupal module limitations.

rails sync:events[events.xml]

Still fails on image but that is a separate issue.


app/views/events/_events.html.erb
```diff
@@ -2,34 +2,39 @@
...
-          ext: <%= event.external_space %>
+          | <%= event.external_building %>
         <% end %>
       </strong></p>
       <p><%= sanitize strip_tags(event.description.html_safe.truncate(250)) %></p>
       <p>
         <% unless event.get_types.nil? %>
           <% event.get_types.each do |type| %>
-            <%= link_to type, event, class: "event_type" %> 
+            <% unless request.env['PATH_INFO'].include?("past") %>
```
bibliotechy: Rails has a nice helper for this
```suggestion
            <% unless action_name == "past" %>
```

config/routes.rb
```diff
@@ -63,18 +63,14 @@
...
-  #
-  # If a finding aid can belong to multiple collections, we can't
-  # determine which collection to display as the parent collection
+  controller :collections do
+    get "finding_aids/:id" => :finding_aids
+  end
 
   controller :events do
-    get "events/past" => :index
+    get "events/past" => :past
```
bibliotechy: If we add a , `as: :past_events` parameter here
```suggestion
    get "events/past" => :past, as: :past_events
```

we can then replace all  calls to the controller and action with "past_events_path". See comments inline above.

db/schema.rb
```diff
@@ -182,8 +182,8 @@
     t.integer "person_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["finding_aid_id"], name: "index_finding_aid_responsibilities_on_finding_aid_id"
```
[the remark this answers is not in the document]
bibliotechy: @cdoyle-temple this one is still outstanding and needs to be addressed as it will cause an issue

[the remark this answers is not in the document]
cdoyle-temple: This file is no longer in the commit, not sure why this is still showing up here.

app/views/events/_events.html.erb
```diff
@@ -2,36 +2,41 @@
...
       </strong></p>
       <p><%= sanitize strip_tags(event.description.html_safe.truncate(250)) %></p>
       <p>
         <% unless event.get_types.nil? %>
           <% event.get_types.each do |type| %>
-            <%= link_to type, event, class: "event_type" %> 
+            <% unless action_name == "past" %>
+              <%= link_to type, events_path(type: type), class: "event_type" %> 
+              <% else %>
+                <%= link_to type, {controller: :events, action: :past, type: type}, class: "event_type" %>
```
bibliotechy:
```suggestion
                <%= link_to type, past_events_path(type: type), class: "event_type" %> 
```

PR #379, opened 2019-04-12
MAN-214-Add UI to see version history in Admin edit page (rebase)

nomadicoder:
- Shows history at bottom of form below submit button

Display and revert history

- Display version history on edit form
- Allow user to select a version to revert to
- Display differences for fields that changed
- Chop trailing end of line characters so we don't get a new version
  when we edit a non-description field, because editor adds in extra
  end-of-line characters when we update a record

Implement button with action

- Implement with button action.
- Localize labels

Make revert method reusable

- Put revert action in application_controller
- Provide revert action for applicable models

Refactor and add versions to item show page

- Add the version list to an item's show page
- Move version template into their own partials in case we need
  to customize an entity's show or edit page.


bibliotechy:
So...the controls did not do what I expected.

I'd like to review this in person Monday.

bibliotechy:
Changes to implementation

- [ ] Version listed in reverse chronological order (most recent first)
- [ ] Versions in listing should have three columns, Date (Librarian Readbable), Changes, "Open in Editor" Button
- [ ] Set table td width percentages so they remain proportional
- [ ] When user clicks Open In Editor button, it should open an edit form with the content from the specified version in the form. A user can then save as normal to revert to that version. 
- [ ] Version listing should only be on Show page, not edit page.

config/routes.rb
```diff
@@ -22,6 +22,14 @@
     resources :spaces
     resources :categories
 
+    resources :alerts, :blogs, :blog_posts, :buildings, :collections,
+      :events, :exhibitions, :groups, :finding_aids , :highlights,
+      :people, :policies, :services, :spaces do
+      member do
+        post :revert
```
bibliotechy: Does this still need to exist? Aren't we just using the edit route now?

nomadicoder: Revert method removed.

bibliotechy:
@skng5 #403 implements the changes we talked about.

PR #387, opened 2019-04-19
MAN-261 Create categories dashboard

nomadicoder:


app/fields/icon_field.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
bibliotechy: I don't think this is used? Or am I wrong? A `PhotoField` is used in the dashboard definition, so I'm not sure this would get called anywhere.

nomadicoder: @bibliotechy I wasn't sure about that, since we use the term Icon vs Photo. Do we need to draw the distinction, and will it cause problems if we make Icon a PhotoField?

nomadicoder: IconField delete

PR #398, opened 2019-04-25
Use event address when matching buildings

bibliotechy:
Add the event address into the matching algorithm to lower the chance of
false positives in buidling matches.


bibliotechy:
@cdoyle-temple I actually found a way to improve the FuzzyFind without adding an "archived" field to building model. I would love your review on this when you get a chance.

bibliotechy:
@cdoyle-temple Might be worth checking if it handles the MAN-258 issues as well.

PR #403, opened 2019-04-30
Use reify instead of custom calls.

bibliotechy:
Also switch to a before_action filter instead of overriding edit.

Remove initializers to serialize as JSON.


bibliotechy:
@skng5 please review

PR #426, opened 2019-05-13
#MAN-283 code services entity page

cdoyle-temple:
#MAN-283 

***********************

Updates/additions from ticket after initial merge.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,12 +21,19 @@
 //= require_tree .
 
 function toggle(x) {
-  if (x instanceof Object) {
-    x.classList.toggle("change");
+  // if (x instanceof Object) {
+  //   x.classList.toggle("change");
+  // }
+  // else {
+  //   document.getElementById(x).classList.toggle("change");
+  // }
+  if (x == "secondary") {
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/application.js
```diff
@@ -21,12 +21,19 @@
 //= require_tree .
 
 function toggle(x) {
-  if (x instanceof Object) {
-    x.classList.toggle("change");
+  // if (x instanceof Object) {
```
bibliotechy: Is there a reason not to remove this commented out code?

bibliotechy: @cdoyle-temple fyi :point_up_2:

cdoyle-temple: I as thinking ahead to the refactoring (event listener) I'll be doing and wanted a history of what I'd done. It was just a quick fix to avoid a little bug.

cdoyle-temple: I'll remove it if you would prefer.

bibliotechy: @cdoyle-temple :man_shrugging: 
If you think it'll be cleaned up soon in another PR, then I'm fine leaving it

PR #427, opened 2019-05-13
MAN-282 External link entity

nomadicoder:
Issue #MAN-282

- Add external_link entity
- Add external_link dashboard
- Make categorizable

### DIFF
--- a/app/dashboards/external_link_dashboard.rb
+++ b/app/dashboards/external_link_dashboard.rb
@@ -31,7 +31,6 @@ class ExternalLinkDashboard < BaseDashboard
   # SHOW_PAGE_ATTRIBUTES
   # an array of attributes that will be displayed on the model's show page.
   SHOW_PAGE_ATTRIBUTES = [
-    :versions,
     :id,
     :title,
     :link,
--- a/app/dashboards/external_link_dashboard.rb
+++ b/app/dashboards/external_link_dashboard.rb
@@ -2,7 +2,7 @@
 
 require "administrate/base_dashboard"
 
-class ExternalLinkDashboard < Administrate::BaseDashboard
+class ExternalLinkDashboard < BaseDashboard
   # ATTRIBUTE_TYPES
   # a hash that describes the type of each of the model's fields.
   #
@@ -10,7 +10,6 @@ class ExternalLinkDashboard < Administrate::BaseDashboard
   # which determines how the attribute is displayed
   # on pages throughout the dashboard.
   ATTRIBUTE_TYPES = {
-    versions: Field::HasMany.with_options(class_name: "PaperTrail::Version"),
     id: Field::Number,
     title: Field::String,
     link: Field::String,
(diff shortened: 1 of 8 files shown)


app/dashboards/external_link_dashboard.rb
```diff
@@ -0,0 +1,61 @@
...
+
+class ExternalLinkDashboard < Administrate::BaseDashboard
+  # ATTRIBUTE_TYPES
+  # a hash that describes the type of each of the model's fields.
+  #
+  # Each different type represents an Administrate::Field object,
+  # which determines how the attribute is displayed
+  # on pages throughout the dashboard.
+  ATTRIBUTE_TYPES = {
+    versions: Field::HasMany.with_options(class_name: "PaperTrail::Version"),
```
bibliotechy: We don't seem to be doing this in any other dashbaord, so I think this is out of place.

app/dashboards/external_link_dashboard.rb
```diff
@@ -0,0 +1,61 @@
...
+  COLLECTION_ATTRIBUTES = [
+    :id,
+    :title,
+    :link,
+  ].freeze
+
+  # SHOW_PAGE_ATTRIBUTES
+  # an array of attributes that will be displayed on the model's show page.
+  SHOW_PAGE_ATTRIBUTES = [
+    :versions,
```
bibliotechy: We don't seem to be doing this in any other dashbaord, so I think this is out of place.

app/dashboards/external_link_dashboard.rb
nomadicoder: - Removed.
- Also changed the class the dashboard inherits from.

app/dashboards/external_link_dashboard.rb
nomadicoder: - Removed.
 -  Also changed the class the dashboard inherits from.

app/dashboards/external_link_dashboard.rb
bibliotechy: I still see this line

app/dashboards/external_link_dashboard.rb
bibliotechy: @skng5 I still see this line.

app/dashboards/external_link_dashboard.rb
bibliotechy: @skng5 :point_up:

nomadicoder: I see it too. I missed a step pushing it to GitHub. It should be there now.

app/dashboards/external_link_dashboard.rb
bibliotechy: @skng5 I still see this line

bibliotechy: Which would now cause an error because the attribute is not defined for the dashboard.

PR #434, opened 2019-05-15
MAN-242 Include Updated Date in JSON response

nomadicoder:


app/serializers/blog_serializer.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class BlogSerializer
+class BlogSerializer < ApplicationSerializer
   include FastJsonapi::ObjectSerializer
```
bibliotechy: If `include FastJsonapi::ObjectSerializer` is in the super class, it doesn't need to be in all of the subclasses.

app/serializers/event_serializer.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class EventSerializer
+class EventSerializer < ApplicationSerializer # ~> NameError: uninitialized constant ApplicationSerializer
```
bibliotechy: What's this error message?

app/schemas/blog_schema.json
```diff
@@ -112,6 +112,16 @@
 								true,
                 false
 							]
+						},
+						"updated_at": {
+							"$id": "#/properties/data/properties/attributes/properties/updated_at",
+							"type": "string",
+							"title": "The time the blog record was updated",
+              "format": "date-time",
```
bibliotechy: what's up with the weird indentation?

app/serializers/event_serializer.rb
nomadicoder: @bibliotechy - That's from the Seeing Is Believing Vim plugin to execute ruby from within Vim.  That was generated by an errant keystroke. Please flag them if you see them. They should not be in the code.

PR #445, opened 2019-05-20
#MAN-251 Events sync service...

cdoyle-temple:
Update non-working facets. 
Fix styles broken from new header design.

Once this is merged, I will delete the current events and import new events from the Drupal export. The test on this locally works as expected.

### DIFF
--- a/app/assets/stylesheets/events.scss
+++ b/app/assets/stylesheets/events.scss
@@ -1,80 +1,64 @@
 @import "partials/_vars.scss";
 
-/* HOMEPAGE */
-.ctaRow.events {
-	a {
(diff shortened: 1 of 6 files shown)


cdoyle-temple:
@nomadicoder This is the same logic that passed previously. I thought it was not a good idea to build the html in the controller?

nomadicoder:
The logic should go in the model. The field we're displaying in the view with the complex logic should be in a method and that method should be called in the view.

nomadicoder:
I see what's going on now. The changes in the actual styling, not just in the what's being displayed.

PR #458, opened 2019-05-24
#MAN-256 Building entity - rename hours field

cdoyle-temple:
Updated label on form field to Hours Identifier, 
made it admin only 
and did the same to spaces and services.

Removed unused custom field partials for hours field
updated dashboards to not show hours field other than on form page
updated tests and serializers to not check for hours field
hours field is just a text identifier the hours controller uses to associate a building/service/space with a column in a google spreadsheet.


cdoyle-temple:
Weird. Changed it and pushed again.

app/views/fields/hours_field/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="field-unit__label">
-  <label for="building_hours">Slug</label>
+  <label for="building_hours">Hours Identifier</label>
```
bibliotechy: @cdoyle-temple We should follow the [administrate practice](https://github.com/thoughtbot/administrate/blob/master/app/views/fields/has_many/_form.html.erb#L23) of dynamically defining this label otherwise, it will be `label for="building_hours"` even in the spaces form, which would be incorrect. 
```suggestion
  <%= f.label field.attribute, for: "#{f.object_name}_#{field.attribute_key}" %>
```
It does mean you will have to define a translation for these forms as well. But it will provide flexibility that they could be different across forms.

PR #463, opened 2019-05-28
#Man-346 hours display for buildings and spaces

cdoyle-temple:
Please create the ability to display relevant hours in both the building and space entities. The design for the building page is in MAN‌-344 . Jackie Sipes is still working on the wireframe for the space entity display. However, it should look the same as the building display page.

This should be a widget/helper that can be dropped into any entity display page that has hours tracked in the spreadsheet.

****************************************

Add hours logic to to concern, has_hours
create model partial for hours display, currently building only
Extended formatting not included in this ticket as it will be applied through the model's views.

### DIFF
--- a/app/controllers/library_hours_controller.rb
+++ b/app/controllers/library_hours_controller.rb
@@ -29,38 +29,7 @@ class LibraryHoursController < ApplicationController
   end
 
   def buildings
-    @buildings = [
-      {
-        slug: "ambler",
-        spaces: ["ambler"]
-      },
-      {
-        slug: "blockson",
-        spaces: ["blockson"]
-      },
-      {
-        slug: "charles",
-        spaces: [
-                  "charles",
-                  "service_zone",
-                  "cafe",
-                  "scrc",
-                  "scholars_studio",
-                  "success_center",
-                  "ask_a_librarian",
-                  "asrs",
-                  "guest_computers"
-                ]
-      },
-      {
-        slug: "ginsburg",
-        spaces: ["ginsburg", "innovation"]
-      },
-      {
-        slug: "podiatry",
-        spaces: ["podiatry"]
-      }
-    ]
+    @buildings = Rails.configuration.building_spaces
   end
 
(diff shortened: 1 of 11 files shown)


cdoyle-temple:
@bibliotechy should be good now

app/controllers/library_hours_controller.rb
```diff
@@ -1,10 +1,43 @@
...
+    else
+      @monday = @today.beginning_of_week
+      @sunday = @today.end_of_week
+      @next_week = @today.next_week
+      @last_week = @today.prev_week
+    end
+  end
+
+  def buildings
     @buildings = [
```
nomadicoder: @cdoyle-temple The `@buildings` is really a configuration and should be pulled in from a locale specific file kept in the `config` directory.  Let's talk about this one.

app/controllers/library_hours_controller.rb
```diff
@@ -1,10 +1,43 @@
...
 
   def index
+    @buildings.each do |building|
+      building[:spaces].map! do |space|
+        space = [building[:slug], LibraryHour.where(location_id: space, date: @monday..@sunday + 1)]
+      end
+    end
+  end
+
+  def set_dates
```
nomadicoder: This logic looks like a setter. I think it should be in `library_hours` model.

app/controllers/library_hours_controller.rb
nomadicoder: This looks like a setter too, but since it's setting attributes of a building, it probably belongs in a concern.

app/controllers/library_hours_controller.rb
nomadicoder: Did some further review. I see these are controller instance variables, so it's okay.  I'll resolve this conversation.

app/controllers/library_hours_controller.rb
nomadicoder: I've created a configuration in the applications config.  These are also controller instance variables, so I think we're good here too.

PR #466, opened 2019-05-30
MAN-278 create custom category route helper

nomadicoder:
Issue #MAN-278

Returns the URL and Path from the first Category instead of the custom URL if it exists.


nomadicoder:
@cdoyle-temple is updating the Jira ticket.

PR #472, opened 2019-06-07
#MAN-360 About Index Page - Links

cdoyle-temple:
Updated links on the About index page to go to first item in the categories displayed.


app/views/pages/visit.html.erb
```diff
@@ -3,37 +3,15 @@
...
-			<p>Lorem ipsum magillicus</p>
-		</div>
-		<div class="col-12 col-md-6 col-lg-4 text-center">
-			<%= link_to (image_tag "study.png", class: "category-icon"), controller: "pages", action: "visit" %>
-			<h2><%= link_to "Study Spaces", controller: "pages", action: "visit" %></h2>
-			<p>Lorem ipsum magillicus</p>
-		</div>
+		<% @categories.each do |category| %>
+					
+					<div class="col-12 col-md-6 col-lg-4 text-center">
```
bibliotechy: In `app/views/pages/about.html.erb` you check for presence of a `category.icon`, but in this template you do not? Seems like we'd always want to check and fall back on a default.

If that is the case, please refactor rendering an individual category into a partial that can be used across all pages what have groups of categories.

cdoyle-temple: This page will be revised in a separate ticket. It's not part of the work done for this ticket as part of the redesign.

cdoyle-temple: This file actually should have no changes for this ticket. Just saw that it does, will revert, and add in the partial for displaying categories. Although, the contact us category will be different as we have to allow for the chat widget.

cdoyle-temple:
@bibliotechy I've created the partial, and since it is such a minor thing now to create an index page for a category, I put the work for the other two pages, Visit the Libraries, and Research Services, into this ticket. I'll close those out in Jira when this one gets closed.

PR #473, opened 2019-06-07
#MAN-359 Update Responsive Menu on Website

cdoyle-temple:
We are going back from the "see more/see less" revision.

- Please go back to the category name appearing instead of "see more"
- DO NOT put the word "menu" on it anywhere
- Remove second instance of the category name (it should only appear once, on the open/close menu heading
- when there are TWO category names, please put an "and" in between them where the open/close category menu names appear (also making sure the category names are removed from anywhere else in the menu)

***************************************

Change responsive menu titles back to using the category names.


app/views/pages/_mobile_menu.html.erb
```diff
@@ -8,7 +8,12 @@
           <div class="bar2"></div>
           <div class="bar3"></div>
         </span>
-        <span id="mobile-navbar-brand" class="navbar-brand">See More</span>
+        <span id="mobile-navbar-brand" class="navbar-brand">
+         <% @page.categories.each do |i| %>
```
bibliotechy: You could simplify this with 
`<%= @categories.collect(&:label).join(" and ") %>`

app/views/services/_mobile_menu.html.erb
```diff
@@ -8,7 +8,12 @@
           <div class="bar2"></div>
           <div class="bar3"></div>
         </span>
-        <span id="mobile-navbar-brand" class="navbar-brand">See More</span>
+        <span id="mobile-navbar-brand" class="navbar-brand">
```
bibliotechy: This seems to exactly duplicate what's above. Could be refactored into a partial or a helper. Something like:
```
def menu_category_list(categories)
  categories.collect(&:label).join(" and ")
end

```

app/views/pages/_mobile_menu.html.erb
cdoyle-temple: nice. done.

app/views/services/_mobile_menu.html.erb
cdoyle-temple: helper. may need a hand on how to test.

bibliotechy: Testing would look like
- create 2 categories with factories
- create another entity
- add the 2 categories to the entity

and then something like:
`expect(helpers.menu_category_list(entity.categories)).to eql "Cat1 and Cat2"`

You probably want to test that it does what yu expect with a single category as well.

https://github.com/tulibraries/manifold/blob/master/spec/helpers/route_helpers_spec.rb#L8-L13

bibliotechy: @cdoyle-temple :point_up:

cdoyle-temple: Rspec test is done, Travis checks have passed.

app/views/pages/_mobile_menu.html.erb
nomadicoder: I see this in app/helpers/application_helper.rb

PR #474, opened 2019-06-07
Parameterize Category#items

bibliotechy:
Adds a `limit_to` param  to Category#items to limit the items
returned to the classes indict`aed in the array.

Adds a Category#child_categories convenience method to just get the
child categories of a category.


cdoyle-temple:
Looks ok to me. We're not currently restricting items in a category to just one model, that I know of, but if we have that need, this would work nicely.

PR #478, opened 2019-06-10
#MAN-344 code building entity page

cdoyle-temple:
Most of this content should match the Services Entity coding.

Building name: H1 bold
"Weekly Hours" H2 bold
Month name: H3 bold
keep 3 column layout in top section. Map and button should take up righthand 1/3 of the page.
The weekly hours should take up 2/3 of the righthand of the page
lefthand menu and green button formatting should match services page
What size should the email/phone number address be? Style-wise they can be the same size as the body copy, but for accessibility I don't know if they should scan as headers. Please proceed as you see fit.

Anything else I'm forgetting?

Please also use your judgment on the size of the map.. I have no idea how long the building descriptions are going to be so it might need to be smaller/shorter.

Also regarding the icons, if we're going to switch to awesome fonts then maybe we can just use placeholders for these for now?

***********************

Updated layout per design
Contact info are not headers for accessibility purposes
Icons can be changed anytime
Moved Hours logic to application controller so hours widget can be used on any page.


app/views/library_hours/show.html.erb
```diff
@@ -1,57 +1,96 @@
...
-    <%= link_to "Home", root_path.to_s %> 
-    <%= link_to "Hours", controller: 'library_hours', action: :index %> 
-    <%= @location.first.name %>
-  </div>
-</div>
-<div class="row">
-  <div class="col">
-    <div class="vertical-stripes-sm">
-      <h2><span><%= @location.first.name %> Hours</span></h2>
+<%# binding.pry %>
```
nomadicoder: Let's take commented out break points out of PR'ed code.

nomadicoder: And please resolve conflicts.

cdoyle-temple: I don't have that binding.pry in my code to remove... Is that an outdated file you were looking at?

nomadicoder: It's gone now. Hmmm.

PR #488, opened 2019-06-13
#MAN-370 code people entity page

cdoyle-temple:
Removed default T image if person has no staff picture.


cdoyle-temple:
Jira request changed again to leave the default T in place for desktop and remove on Mobile, which is how this branch was working before this last change. Closing PR.

PR #499, opened 2019-06-21
Error on linode when external link added to a space

cdoyle-temple:
Show page errors as it can't find external_link_path.

Updated call to entity_path for category items using custom category_url in routes.rb.


cdoyle-temple:
/spaces/21 currently errors.

PR #503, opened 2019-06-24
#MAN-350 Alphabetize links in categories

cdoyle-temple:
Please alphabetize the links that appear in a given category. These links will display in the left hand menu.

***************************

Secondary mobile menu sorted, working on secondary desktop.
Secondary desktop sorted.
pages with menus will error if external link is in category, but that should be fixed in another ticket.


app/models/external_link.rb
```diff
@@ -7,4 +7,8 @@ class ExternalLink < ApplicationRecord
 
   validates :title, presence: true
   validates :link, presence: true, url: true
+
+  def label
```
bibliotechy: The preferred method for adding a label is via the `label_method`, as defined in the parent model https://github.com/tulibraries/manifold/blob/master/app/models/application_record.rb#L12
```suggestion
  def label_method
```

app/views/buildings/_categories.html.erb
```diff
@@ -11,7 +11,8 @@
     <div class="index-items">
       <div>
         <ul class="list-unstyled">
-          <% cat.items.each do |c| %>
+          <% cat.items.sort_by{|e| e.label}.each do |c| %>
+          <% unless c.class.to_s == "Category" %>
```
bibliotechy: Once #505 is merged, you can use the `items(exclude: [:category])` syntax

app/models/external_link.rb
cdoyle-temple: @bibliotechy It appears this way in all the other models, should I change them as well?

app/views/buildings/_categories.html.erb
cdoyle-temple: Should this be resolved in the new ticket for the re-use of code (shared partial/concern...)

bibliotechy: No, I feel like this would benefit from rebasing this branch against master  and then applying this change, and then force pushing

app/models/external_link.rb
bibliotechy: Hmm, you are right. No, you can disregard this comment

app/views/buildings/_categories.html.erb
cdoyle-temple: Done.

bibliotechy:
@cdoyle-temple I've made a small PR against this branch with a few more things to clean up , #515 . Review and merge that and then we can megre this.

PR #506, opened 2019-06-24
#MAN-394 Code Collections Page

cdoyle-temple:
Technical Requirements: 

*Left column*: 
Menu - pull in the category menu that the collection belongs to. 

*Middle column*: 
Image
About - text from "Description" field
Subjects - list the subjects as text, not linked to anything at this point.
Location - we current have a field in the model for space 
- the text/link with the map icon should go to the related space
- under contact, if possible pull in the phone number and email address from the related space

*Right Column*: 

Please use the top green button (our standard "Call to action" button)
- Text should read "Explore this collection"
- Destination: all of the finding aids that are tagged with this particular collection - I know that we don't have the related finding aids in the collection model, but are you able to do some backward engineering to find all the finding aids that are tagged with this collection?

Remove the "link to related policy"
Remove the other "explore this collection" box
Remove the "Contact" box from the right 

***************************

Updated collection show page according to ticket.


cdoyle-temple:
> @cdoyle-temple It feels like there are major duplications with #502 that was just merged. Are there ways we can reuse code across both contexts?

Yes, perhaps we could open a separate ticket for that refactoring? It's hard to see the possible places for that until you've done one or two of these, and then it becomes more apparent.

bibliotechy:
@cdoyle-temple that makes sense. https://tulibdev.atlassian.net/browse/MAN-404

PR #516, opened 2019-06-27
#MAN-349 Code Staff Directory page

cdoyle-temple:
Design should match the events feed.. mostly.

From ticket, Jackie:

Yes, I agree that the staff type filter should be open by default. 

I still think that filter functionality should the same as library search: 
-one selection per filter (e.g. once a user selects a "location," results are narrowed and filter values for "subject specialty" and "library department" are automatically narrowed to reflect any remaining options – same as in Library Search) 
-the user should be able to select one value from each filter (library location, subject, department), but we do not need to provide the option of applying multiple limits within each filter

so... that would mean no submit button and we'd need a "reset filters" start over button.

*********

Matched given design, developed filters to narrow available filters as user drills down.
Note: Removed All Staff filter due to the Close button (x) on Specialists Only making that selection redundant.


app/controllers/concerns/person_filters.rb
```diff
@@ -0,0 +1,41 @@
...
+  def specialties_list(persons)
+    to_add = persons.select { |person| person.specialties.try(:any?) { |subject| subject.include?(params[:specialty]) }  }
+
+    special_people = []
+
+    to_add.each do |person|
+      special_people << person
+    end
+
+    special_people
```
bibliotechy: lol

PR #521, opened 2019-07-01
#MAN-383 add external link field to collection entity

cdoyle-temple:
Please add the external link field to the collection entity

********************

Added link field to model, dashboard and view.


cdoyle-temple:
Added test on external link for collection model.

PR #556, opened 2019-07-12
MAN-266 detach attachments

nomadicoder:
Addresses Issue #MAN-266


nomadicoder:
@bibliotechy Added post detach action to handle attaching a file when creating a new object. Should fix the missing action error.

PR #561, opened 2019-07-15
Man 416 code the homepage

cdoyle-temple:
Fixes the instance variable print my paper not being displayed


bibliotechy:
@cdoyle-temple I think this PR would merge cleanly and be a little easier to understand if you created a new branch based off qa with your new change

PR #568, opened 2019-07-15
#MAN-363 Contact us Index page links

cdoyle-temple:
Please link the “Frequently called numbers” to https://web.qa.tul-infra.page/pages/19

Please link the “Find your subject librarian” to https://web.qa.tul-infra.page/people?page=1&specialists=true


app/controllers/pages_controller.rb
```diff
@@ -83,6 +80,7 @@ def index
   end
 
   def contact
+    @fcn_link = Page.find_by_title("Frequently called numbers")
```
bibliotechy: @cdoyle-temple Now that we have a slug on this model, can we update the page and query by slug?

PR #570, opened 2019-07-16
#MAN-433 Staff directory and email not showing in responsive view

cdoyle-temple:
The icons/menu items for both the staff directory and the email function are not showing in the responsive view

***************************

Updated page to include missing partials in responsive view.


bibliotechy:
@cdoyle-temple can you please resolve these conflicts, as I'm not sure which value is the correct one

PR #575, opened 2019-07-16
#MAN-419 On staff directory page, add link to Frequently called numbers

cdoyle-temple:
On the Staff directory, there is a call-out box for "Frequently called numbers". That box should link to https://web.qa.tul-infra.page/pages/19.

Also, please update the “Main phone” to read 215-204-0744

### DIFF
--- a/app/controllers/persons_controller.rb
+++ b/app/controllers/persons_controller.rb
@@ -8,6 +8,7 @@ class PersonsController < ApplicationController
 
   def index
     @persons = Person.all
+    @fcn_link = Page.find_by_title("Frequently called numbers")
     respond_to do |format|
       format.html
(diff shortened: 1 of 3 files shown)


app/controllers/persons_controller.rb
```diff
@@ -8,6 +8,7 @@ class PersonsController < ApplicationController
 
   def index
     @persons = Person.all
+    @fcn_link = Page.find_by_title("Frequently called numbers")
```
bibliotechy: @cdoyle-temple Now that we have slugs on this model, do you want to update this page with a slug and query by slug instead?

cdoyle-temple: I was going to save all that until these mini pages are done, and then go back and fill them all in in. But it's your call. Let me know.

PR #579, opened 2019-07-17
Events feed from Drupal

cdoyle-temple:
Added logic to filter helper where locations are not supplied from old events in Drupal to ignore the nils.

### DIFF
--- a/app/controllers/concerns/event_filters.rb
+++ b/app/controllers/concerns/event_filters.rb
@@ -23,7 +23,6 @@ module EventFilters
         end
       end }
 
-    locations.reject! { |s| s.nil? }
-    locations.uniq.sort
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/event_filters.rb
```diff
@@ -18,9 +18,12 @@ def locations_list(events)
       unless event.building.nil?
         event.building.label
       else
-        event.external_building
+        unless event.external_building.nil?
+          event.external_building
+        end
       end }
 
+    locations.reject! { |s| s.nil? }
```
bibliotechy: `locations.compact` will eliminate `nil`s more efficiently.

You could just add it into the next line as `locations.compact.uniq.sort`

cdoyle-temple: nice

PR #602, opened 2019-07-23
MAN-460 Don't process show image if size matches show image dimensions

nomadicoder:
- Addresses Issue #MAN-460
- Fixes error thrown if image dimensions match show image dimensions,
  If the dimensions match, mogrify optimization throws error due to
  failure to find expected temporary file to mogrify in /tmp directory


nomadicoder:
@cdoyle-temple Good question. Are we assured that all images will will be for show view? What if icons are thumbnail sized?

PR #626, opened 2019-07-26
#MAN-389 Modify color of top nav to show page

cdoyle-temple:
We want the top navigation to be highlighted when you're on one of those pages-- About, Visit & Study, etc... 

As a result, please make all of the navigation items the same green with the darker color vertical line separating each item. Then, when a user is on the top-level index page (About, Visit & Study, etc), that nav item will be highlighted using the darker green. However, when the user navigates to a sub-index page  from the top-level index page, all nav goes back to regular green. 

Remove all vertical lines in primary navigation.
*******************
Added checks for active page and highlighted corresponding menu item.


nomadicoder:
It appears that green is actually black.

PR #650, opened 2019-07-31
MAN-509 Form elements must have labels

cdoyle-temple:
Added labels to Search everything forms not generated by Rails.
Added style to hide labels but keep available for screen readers.

### DIFF
--- a/app/views/application/_mobile-search.html.erb
+++ b/app/views/application/_mobile-search.html.erb
@@ -4,8 +4,9 @@
     <!-- <div class="padding"></div> -->
     <div class="navbar-nav">
       <form action="https://librarysearch.temple.edu/bento" class="mobile">
+        <label for="q" class="visuallyhidden">Search Everything</label>
         <input type="text" name="q" id="librarySearch" class="mobile" placeholder="Search everything" autocomplete="off" />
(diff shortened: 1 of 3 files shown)


app/views/application/_mobile-search.html.erb
```diff
@@ -4,8 +4,9 @@
     <!-- <div class="padding"></div> -->
     <div class="navbar-nav">
       <form action="https://librarysearch.temple.edu/bento" class="mobile">
+        <label for="q" class="visuallyhidden">Search Everything</label>
         <input type="text" name="q" id="librarySearch" class="mobile" placeholder="Search everything" autocomplete="off" />
-        <input type="submit" class="mobile" value = "">
+        <input type="submit" class="mobile" value = "" aria-label="submit search query">
```
bibliotechy: @cdoyle-temple Why is the first input using a `visuallyhidden` label tag, but the second  uses an inline `aria-label` attribute? Not criticism, but asking what the difference is?

cdoyle-temple: The first is a label for the search field itself. We can hide that as it isn't part of the design. The second is a label for the submit button which we can't really hide.

PR #678, opened 2019-08-05
Category Icon Fix

cdoyle-temple:
Add imageable concern to Category model. 
Remove image field from model.

View was not finding image.attachment  before these changes.

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -3,13 +3,12 @@
 class Category < ApplicationRecord
   include Rails.application.routes.url_helpers
   #TODO: should we validate that icon is svg?
+  include Imageable
 
   has_many :categorizations, dependent: :destroy
 
   has_many :nested_categorizations, as: :categorizable, dependent: :destroy, class_name: "Categorization"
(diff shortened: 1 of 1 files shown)


bibliotechy:
@cdoyle-temple Which view? I see them on the [landing pages](https://web.qa.tul-infra.page/research-services). Where should I check this?

cdoyle-temple:
@bibliotechy This was in regard to an issue that the icons were not showing up when a new icon was uploaded for a category. In testing locally, the image.attachment.nil? was coming back true and so skipping the uploaded icon and presenting the default Temple T icon. This fix should allow the image.attachment.nil? to come back false and present the correct icon. You should see the uploaded icons on the About page.

bibliotechy:
@cdoyle-temple But if I go to https://web.qa.tul-infra.page/about right now, I see images.

cdoyle-temple:
So that's unexpected given the conversation this morning with Rachel and Cynthia. One thing this fix does do, though, is make the Category model more like other models that have images by using the imageable concern, which it was not doing.

PR #698, opened 2019-08-07
Add a redirect model and dashboard

bibliotechy:
Adds a redirect model that allows administrate users to

- define an incoming legacy_path ("/scrc/ask")
- define either
  - a static local path it should go to ("/forms/ask-scrc")
  - a redirectable object it should redirect to. Can choose from
    - Buildings
    - Collections
    - Pages
    - Spaces
    - Services
    - Policies

Any non matched redirect reverts to a 404

### Static URLS
- Go to the Redirects  admin dashboard  - /admin/redirects
- Create a new redirect
  - add a legacy_url i.e ("scrc/ask-here") (should not start with a slash)
  - add known static path in the manifold_url ("/forms/ask-scrc") (needs to start with a slash)
- go to localhost:3000/scrc/ask-here - it shoudl redirect to localhost:3000/forms/ask-scrc

### Entity redirects
- Create a new redirect
  - add a legacy_url i.e ("some/place") (should not start with a slash)
  - add a redirectable by selecting an entity from the drop down
- go to localhost:3000/some/place - it should redirect to the entity selected in the drop down

### DIFF
--- a/app/controllers/admin/redirects_controller.rb
+++ b/app/controllers/admin/redirects_controller.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+module Admin
+  class RedirectsController < Admin::ApplicationController
+  end
+end
--- a/app/controllers/redirects_controller.rb
+++ b/app/controllers/redirects_controller.rb
@@ -3,10 +3,11 @@
 class RedirectsController < ApplicationController
   def show
     if params[:path]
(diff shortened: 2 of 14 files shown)


bibliotechy:
@cdoyle-temple I added one change - enforcing uniqueness on the legacy_path value. When that passes, can you please merge?

PR #805, opened 2019-09-10
Order the gems in gemfile

bibliotechy:
So that in the future we can just insert them in alpha order.


bibliotechy:
@nomadicoder done

PR #816, opened 2019-09-11
#MAN-520 Collection entity explore this collection button

cdoyle-temple:
I'm realizing that the Collection entity can only be used for SCRC or blockson collections that have finding aids in the finding aid index. The "explore this collection" appears on every page. This is misleading if we use this entity for collections that don't have finding aids. For the MVP, I'm going to use either the Page entity or the Space entity for non-scrc/blockson collections until we can come up with a better solution. 
******************
Add concern that looks up aids per collection
Display button only if concern method returns true


app/controllers/concerns/finding_aids_finder.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module FindingAidsFinder
+  extend ActiveSupport::Concern
+
+  def has_finding_aids(collection)
+    unless number_of_aids(collection) == 0
```
nomadicoder: Unless there's good reason to use an unless/else block I'd reduce this to one line of code like `number_of_aides(collection) != 0`. This may be all you need in this method.

cdoyle-temple:
@nomadicoder Updated the concern.

cdoyle-temple:
@nomadicoder Resolved.

PR #820, opened 2019-09-12
#MAN-502 Add LibGuides to person profile

cdoyle-temple:
While I was working on a project last week I noticed that on the current list of librarian liaisons, there is no link to a librarian’s subject guides.
Here’s an example for Adam:
https://guides.temple.edu/prf.php?account_id=15120
In checking with the LRS Unit Heads, there was consensus that it would be helpful to provide these links on the profile pages in the new website.
******************
To test: update Adam's entry and add the above id.


cdoyle-temple:
The means of inclusion for libguides is yet to be decided and needs to be further discussed, so this PR as premature.

PR #855, opened 2019-09-26
#MAN-502 add lib guides to person profile

cdoyle-temple:
- Adds libguide account field to model
- when populated, will show button on the person view page linking to springshare page listing that person's libguides.


cdoyle-temple:
Errant migration was throwing this off. It did not belong with this work, so it was removed. (add slug to collections)

PR #859, opened 2019-09-26
#MAN-673 aeon api tweak

cdoyle-temple:
Once the new migration is in place, the slug field for the Blockson Collection collection will need the "blockson_collection" identifier added to it.

Then you will not see the Request this item.... link button on any blockson finding aids.

Find blockson aids here: /finding_aids?collection=10

### DIFF
--- a/app/controllers/finding_aids_controller.rb
+++ b/app/controllers/finding_aids_controller.rb
@@ -53,6 +53,6 @@ class FindingAidsController < ApplicationController
       @finding_aid = FindingAid.find(params[:id])
       @title = @finding_aid.label
       blockson = Collection.find_by_slug("blockson_collection")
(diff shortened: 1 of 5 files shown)


app/controllers/finding_aids_controller.rb
```diff
@@ -51,5 +51,8 @@ def get_filters(aids)
   private
     def set_finding_aid
       @finding_aid = FindingAid.find(params[:id])
+      @title = @finding_aid.label
+      blockson = Collection.find_by_slug("blockson_collection")
+      @aeon = @finding_aid.collections.include?(blockson) ? true : false
```
nomadicoder: You don't need the `? true : false` here, the `include?` method will give you the same result.

PR #861, opened 2019-09-26
#MAN-675 building name helper function can't take nil as param

cdoyle-temple:
Events without locations (nil building and blank external building) were erroring the view since that nil, or blank, was being passed to the filter helper which populates the filters.

Empty locations are now handled so that they pass through without triggering the filter helpers.

Test this here: https://library.temple.edu/events/past?page=99 (if no error, fix is good)


cdoyle-temple:
Will refactor to the get_bldg_name method, but the facets partial was another issue that looked like this: (see https://tulibdev.atlassian.net/browse/MAN-675).

PR #901, opened 2019-10-07
#MAN-502 Add LibGuides to person profile

cdoyle-temple:
Rework MAN-502 as it kept failing in Traivs.


db/migrate/20190926180617_add_slug_field_to_collections.rb
```diff
@@ -1,7 +0,0 @@
-# frozen_string_literal: true
```
bibliotechy: Why is this deleted?.

cdoyle-temple: accident, inadvertant.

bibliotechy: @cdoyle-temple ping

cdoyle-temple: yah ahahha

cdoyle-temple:
@bibliotechy migration restored. Found the error was in the schema where a field created in this work was already in merged code.

PR #907, opened 2019-10-09
#MAN-716 website finding aid creation issue

cdoyle-temple:
Can we require that something have either a subject or a collection? That way, they are still browsable in the finding aids index page?
**************
Added custom validation for either/or condition.


cdoyle-temple:
Tests in.

PR #909, opened 2019-10-09
#MAN-664 Add meaningful page titles to head section of the website

cdoyle-temple:
Might be worth implementing something more general, like using https://github.com/kpumuk/meta-tags so we can get all the sweet sweet seo juice.


app/views/persons/_desktop.html.erb
```diff
@@ -10,7 +10,7 @@
         <% end %>
       </div>
       <div class="col-8">
-        <h1 class="person-name page-title"><%= @person.first_name %> <%= @person.last_name %></h1>
+        <h1 class="person-name page-title"><%= title @person.first_name + " " + @person.last_name %></h1>
```
bibliotechy: Please use string interpolation instead of concatenation
`<%= title "#{@person.first_name}  #{@person.last_name}" %>`

It's more idiomatic ruby, plus it is safer against unexpected bad data like `@person.first_name` being nil.

app/views/persons/_mobile.html.erb
```diff
@@ -3,7 +3,7 @@
     <% if @person.image.attached?  %>
       <%= image_tag @person.show_image, class: "decorative" %>
     <% end %>
-    <h1 class="person-name"><%= @person.first_name %> <%= @person.last_name %></h1>
+    <h1 class="person-name"><%= title @person.first_name + " " + @person.last_name %></h1>
```
bibliotechy: Please use string interpolation instead of concatenation
`<%= title "#{@person.first_name}  #{@person.last_name}" %>`

It's more idiomatic ruby, plus it is safer against unexpected bad data like `@person.first_name` being nil.

cdoyle-temple: got it

app/views/persons/_desktop.html.erb
cdoyle-temple: got it

cdoyle-temple:
@bibliotechy Updated the name format.

PR #911, opened 2019-10-09
Add weighting to items in categories

bibliotechy:
Items in Categories can now be weighted from 1-10, with 1 being most important, 10 being the least (and the default).

Category Item weightings can be assigned in the Admin Category Edit page.

Weighting will affect order in navigation menus, and order in multiple pages where a categories are used to list the items on a page, like the HSL home page research links, visit links, etc.

This also removes the default alphabetical sort that was in place in those pages.


cdoyle-temple:
If someone deletes a number, accidentally leaving the field blank, submits the form, and then tries to edit the form again, that same error is thrown.

PR #956, opened 2019-10-21
Create admingroup model

nomadicoder:
@bibliotechy What else do we need?


nomadicoder:
Moving forward with show page updates and entities

PR #959, opened 2019-10-22
Man 650 add name field to accounts

nomadicoder:
Add name field to account table


nomadicoder:
Add a field for “Group membership”

PR #960, opened 2019-10-22
MAN-650 Add name field to accounts

nomadicoder:
These changes will account for the addition of account groupings and general additional fields to make account management easier.

On the “Add new/edit” page: 

**Add a field for “Group membership”**

- Field should be a drop-drown menu list of the groups available in the “Admin Group model”
- Each account can only have one group affiliation

**Add a field for “Name”**

- with everyone’s email as tuzxxxx@temple.edu, it’s really hard to know who is who. 
- Please add a free text field for “Name” that appears above “email”.

_On the display page:_ 

- Change the title at the top to say “Show [name]” and “Edit [name]” instead of the account number
- Add field for “Name”

_On the list page:_

- Add a column and display the contents of the “Name” field
- Remove the “Email” column *[Not done yet, because existing records don't have anything in the name field. This should be done after we populate all of the existing records with names.]*
- Add a column and display the contents of the “Admin Group” field


app/dashboards/account_dashboard.rb
```diff
@@ -78,7 +87,7 @@ class AccountDashboard < BaseDashboard
   # Overwrite this method to customize how accounts are displayed
   # across all pages of the admin dashboard.
   #
-  # def display_resource(account)
-  #   "Account ##{account.id}"
-  # end
+  def display_resource(account)
+    "Account ##{account.name}"
```
bibliotechy: @nomadicoder Do we want it to say "Account" in front of the name?

PR #965, opened 2019-10-25
MAN-738 Add structured data to buildings

nomadicoder:
Creates structured data for buildings.

Refactor to create the JSON-LD with a method to create the LD hash in the model and converting that into JSON in an Application Helper. Applied this refactor to Events and Buildings.


nomadicoder:
Missing a few fields

nomadicoder:
Added in all fields except hours.

PR #1017, opened 2019-11-14
Draft entities

nomadicoder:
#MAN-215
Allow draft changes of an entity using the Action-Draft Gem.


nomadicoder:
Withdrawing PR to active development.

PR #1087, opened 2020-01-03
#MAN-786 Create form: Group Visit Request Form

cdoyle-temple:
Add group visit form


app/views/forms/group-visit/_personal_info.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= f.input :name, required: true %>
+<%= f.input :email, as: :email, required: true %>
+<%= f.input :phone, as: :tel, label: "Phone Number", required: true %>
```
nomadicoder: @cdoyle-temple Maybe the field should be called :phone_number or the label just phone so we don't have so many literals floating around.

PR #1093, opened 2020-01-08
#MAN-785 Add option to suppress Update Bookmarks message

cdoyle-temple:
Add no_message field to model
Add logic to concern to check for no_message flag and not specify a flash notice if found
Add tests for redirects with flag and without

### DIFF
--- a/app/controllers/concerns/redirect_logic.rb
+++ b/app/controllers/concerns/redirect_logic.rb
@@ -6,10 +6,11 @@ module RedirectLogic
   def redirect_or_404
     redirect = Redirect.find_by(legacy_path: legacy_path)
     if redirect
-      message =
-      "#{request.env['HTTP_HOST']}#{legacy_path} has moved. \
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/redirect_logic.rb
```diff
@@ -6,10 +6,12 @@ module RedirectLogic
...
     redirect = Redirect.find_by(legacy_path: legacy_path)
     if redirect
-      message =
-      "#{request.env['HTTP_HOST']}#{legacy_path} has moved. \
-      Please update bookmarks and links"
-      redirect_to(redirect.path,
+      unless redirect.no_message
+        message =
+        "#{request.env['HTTP_HOST']}#{legacy_path} has moved. \
+        Please update bookmarks and links"
```
nomadicoder: @cdoyle-temple The explicit text should be in the localized.

cdoyle-temple:
String localized.

PR #1122, opened 2020-01-22
FIx for People JSON

cdoyle-temple:
The instance variable change affects the filters working properly so I changed the variable used for the JSON representation and ingest in the catalog.

### DIFF
--- a/app/controllers/persons_controller.rb
+++ b/app/controllers/persons_controller.rb
@@ -8,11 +8,12 @@ class PersonsController < ApplicationController
 
   def index
     @persons = get_persons
+    json_people = Person.all
     @fcn_link = Webpage.find_by(title: "Frequently called numbers")
 
     respond_to do |format|
       format.html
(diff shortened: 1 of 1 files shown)


app/controllers/persons_controller.rb
```diff
@@ -8,11 +8,12 @@ class PersonsController < ApplicationController
 
   def index
     @persons = get_persons
+    json_people = Person.all
     @fcn_link = Webpage.find_by(title: "Frequently called numbers")
 
     respond_to do |format|
       format.html
-      format.json { render json: PersonSerializer.new(@persons.to_a) }
+      format.json { render json: PersonSerializer.new(json_people.to_a) }
```
dkinzer: @cdoyle-temple I think it would be more efficient to not set `json_people = Person.all` and instead just do 
```
format.json { render json: PersonSerializer.new(json_people.to_a) }
```
Reason being that currently it loads `Person.all` even in cases where you do just want to display a limit of 20 (i.e. format.html).

PR #1128, opened 2020-01-24
#MAN-724 Create a file system model

cdoyle-temple:
Adds uploads to groups, policies, services, spaces, and webpages. PDFs only.

Note: When testing, uploading the wrong file type will throw an error because we can only delete from AWS in production.


cdoyle-temple:
This work needs to be combined with the work from another ticket: #MAN-762

nomadicoder:
@cdoyle-temple Please resolve the conficts.

PR #1151, opened 2020-02-10
#MAN-808 Repeat events on library website

cdoyle-temple:
Further extend event matching by adding start time time stamp to record hash.

### DIFF
--- a/app/services/sync_service/events.rb
+++ b/app/services/sync_service/events.rb
@@ -53,6 +53,7 @@ class SyncService::Events
       "registration_status" => event.fetch("RegistrationStatus", nil),
       "registration_link"   => event.fetch("ExternalRegistrationURL", nil),
       "start_time"          => start_time(event),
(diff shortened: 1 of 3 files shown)


app/services/sync_service/events.rb
```diff
@@ -65,7 +66,7 @@ def record_hash(event)
   def create_or_update_if_needed!(record_hash)
     # If a record already exists with this content hash, then no update needed
     if should_create_or_update(record_hash)
-      event = Event.find_by(content_hash: record_hash["content_hash"]) || Event.find_by(guid: record_hash["guid"])
+      event = Event.find_by(content_hash: record_hash["content_hash"]) || Event.find_by(guid: record_hash["guid"], timestamp_start: record_hash["timestamp_start"])
```
bibliotechy: @cdoyle-temple Using this logic, if someone changes the time of a non-repeating event for typical reasons, like they made a typo or they are rescheduling it for the following day, that will create a new event instance, so there will be vents for both the original time and the new time

PR #1192, opened 2020-02-27
#MAN-816 Add generic language to categories that don't have a long description

cdoyle-temple:
Add default text to Category show page.

### DIFF
--- a/app/views/categories/show.html.erb
+++ b/app/views/categories/show.html.erb
@@ -13,7 +13,11 @@
     <% end %>
         <h1 class="page-title"><%= title @category.label %></h1>
         <div class="details">
-          <%= sanitize @category.long_description.html_safe unless @category.long_description.nil? %>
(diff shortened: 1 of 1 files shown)


app/views/categories/show.html.erb
```diff
@@ -13,7 +13,11 @@
     <% end %>
         <h1 class="page-title"><%= title @category.label %></h1>
         <div class="details">
-          <%= sanitize @category.long_description.html_safe unless @category.long_description.nil? %>
+          <% unless @category.long_description.nil? %>
+            <%= sanitize @category.long_description.html_safe %>
+          <% else %>
+            <p>Page under construction, please check back soon or use the menu on the left to navigate the pages in this section.</p>
```
nomadicoder: We should localize this.

cdoyle-temple: It's pretty specific to categories, and temporary, are you sure you want this in config?

PR #1209, opened 2020-03-03
Use Sitemap-to-s3 gem extension

nomadicoder:
Address MAN-818.

Once this is deployed, we need to add a manifold managed redirect from /sitemap.xml.gz to https://tulib-manifold-prod.s3.amazonaws.com/sitemap.xml.gz, and then this should just work.

Because of the shared bucket, we only want this running in production in airflow, so the stage and qa versions should be turned off.


nomadicoder:
I am unable to access S3 buckets from my local development environment.

nomadicoder:
Sitemap-S3 generator docs: https://github.com/kjvarga/sitemap_generator#sitemapgeneratorawssdkadapter

bibliotechy:
Once https://github.com/tulibraries/manifold/pull/1213 is merged in, this should be good to go.

PR #1219, opened 2020-03-10
Serialization testing

cdoyle-temple:
Update link_serializable concern to better handle object params, and account for pluralization of people in controller names.


nomadicoder:
We should return to this later and see if we can alias persons and people.

PR #1253, opened 2020-03-30
Redirects Fix

cdoyle-temple:
Since we are now using redirects for internal purposes and not just for old Drupal paths, the logic calling the redirects needed to be extended to all models that could be used as a redirect request.


cdoyle-temple:
Still need to update testing.

cdoyle-temple:
Testing updated

PR #1306, opened 2020-04-29
#MAN-898 Update controller links to use external links

cdoyle-temple:
Add redirects to replace the instance variable in the footer and homepage.

This uses a combination of the external links and the internal redirectable types.

New redirects will need to be added for this to be fully functional, and I'l update this description when they are done.


cdoyle-temple:
Redirects in place on QA, ready for merge.

bibliotechy:
💋 👌

PR #1318, opened 2020-05-11
Move dashboard collections into it's own file

nomadicoder:
Issue: #MAN-125

Note - This is a change from the original request, which would have the collections defined in a YAML file. This PR moves the dashboard collections long array definitions into a separate initializer.


bibliotechy:
@nomadicoder seems legit to me. You can write a spec for it if you are nervous. There is one initializer spec as an example, which is pretty simple.

PR #1324, opened 2020-05-12
#MAN -918 Update link for See All libraries and locations

cdoyle-temple:
Update link to use redirect rather than instance variable. 

This will allow admins to change the target without having to request code changes! 

Extend cleanup of instance variables on homepage. Also, extended cleanup to mini pages.


cdoyle-temple:
Needs redirects in place on servers before merge.

cdoyle-temple:
Redirects in place on QA. Before next release, production will need the new redirects as well.

PR #1348, opened 2020-05-26
Docker dev environment

nomadicoder:
For development only. This branch should not affect production.

This branch allows us to develop in a docker environment. It has two containers. One for the Manifold Web application and another for the Postgresql database. Instructions are in docker-readme.md.

Warning - It can take up to two hours to build the Manifold Web Docker image.


entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+# Remove a potentially pre-existing server.pid for Rails.
+rm -f /myapp/tmp/pids/server.pid
```
bibliotechy, line 5: Is this copypasta?

nomadicoder: Mannacotti - Thanks for the catch.

nomadicoder: Changed to /manifold.  Also addressed some S3 and Google API issues, and created a script to clean up and restore the database from data dump.

PR #1448, opened 2020-07-24
Handle nil Finding aid undefined subject search

nomadicoder:
- Also corrects restore database command in docker readme


nomadicoder:
Work already done

PR #1562, opened 2020-10-01
Redirect Internal Instances

cdoyle-temple:
Removes unless @model from controller call to redirect_or_404
Passes @model to redirect_or_404 for processing
Returns redirection to redirect_path for external and internal paths
Rejects not existent internal paths without redirects

### DIFF
--- a/app/controllers/concerns/redirect_logic.rb
+++ b/app/controllers/concerns/redirect_logic.rb
@@ -3,7 +3,8 @@
 module RedirectLogic
   extend ActiveSupport::Concern
 
-  def redirect_or_404
+  def redirect_or_404(instance = nil)
+    # binding.pry
     redirect = Redirect.find_by(legacy_path: legacy_path)
     if redirect
       unless redirect.no_message
@@ -15,7 +16,15 @@ module RedirectLogic
                   notice: message
                   )
     else
-      raise ActionController::RoutingError.new("Not Found")
+      if Rails.application.routes.recognize_path(legacy_path)
+        if instance
+          return
+        else
+          raise ActionController::RoutingError.new("Not Found")
+        end
+      else
+        raise ActionController::RoutingError.new("Not Found")
+      end
     end
   end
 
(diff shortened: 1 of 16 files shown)


bibliotechy:
@cdoyle-temple Can you give me a little more context what the goal of this PR is? It's a bit too abstract for me.

cdoyle-temple:
@bibliotechy The desire is to create redirects that would redirect from a known path/id. Currently, the redirect_or_404 method is restricted to requests that do not set a model instance for the method's show controller, i.e., "redirect_or_404 unless @service".

This prevented creating a "working" redirect from, say, "/services/29" to "/categories/research-data-services" since service.id == 29 exists and so the controller bypassed the redirect_or_404 method entirely.

This work removes the "unless @model" from the redirect_or_404 call in the controller so that the above example will work. It does so by passing the model instance variable to the method rather than blocking the method, allowing the method to check for a redirect with that instance's path as the legacy path.

This work accommodates requests for both found and not found instances with registered paths, and random non-registered paths. All tests passing.

app/controllers/concerns/redirect_logic.rb
```diff
@@ -3,7 +3,7 @@
 module RedirectLogic
   extend ActiveSupport::Concern
 
-  def redirect_or_404
+  def redirect_or_404(*args)
```
bibliotechy, line 6: What's the use case for the splat here? Looks like you only even have one parmaters passed, the instance object or nil. Just kind of obscures what is happening. And then you have to use `args[0]` below, which is kind of weird.

cdoyle-temple: When the controller did not set an instance, an error was thrown in the method when it tried to check the attribute of nil. I needed an optional argument that could also handle nil. Maybe a better way?

bibliotechy: Maybe `def redirect_or_404(object_or_nil=nil)`? 

Providing a default value means it can handle no params.

bibliotechy: @cdoyle-temple Or something like that.

bibliotechy (approved):

PR #1639, opened 2020-12-03
Upgrade Ruby to version 2.7.2

nomadicoder:


sensei100 (changes requested):
@nomadicoder Can you also update the target ruby version in the rubicon.yml file?

sensei100 (approved):

PR #1745, opened 2021-01-28
Pin omniauth to 1 9 1

nomadicoder:
Current version of google-

Created an [issue](https://tulibdev.atlassian.net/browse/MAN-1047) to remind us to check if omniauth-google-auth2 has been upgraded.  Dependabot should catch this for us but we should check ourselves periodically.


nomadicoder:
Base this on qa, not main

PR #1853, opened 2021-03-25
Upgrade to OAuth 2.0

nomadicoder:
- Update Devise to non-released head of master
- Upgrade to OAuth2 version of omniauth-google-oauth2


nomadicoder:
Addressed in https://github.com/tulibraries/manifold/pull/1903

PR #1876, opened 2021-04-06
#MAN-1057 Update to Blockson mini homepage

cdoyle-temple:
Account for blockson category name change.
Update template to check layout rather than name.


dkinzer (approved):
I'm approving this, but I have no way to know that it's working without any tests.

PR #1908, opened 2021-05-05
Fixes OAuth passthru error on non-development env

nomadicoder:
Follows Migrating in Rails Step 4 recommendation in https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284#mitigating-in-rails-applications


nomadicoder:
It looks like we're still getting GET requests, though the code is in place to make POST requests.  
Strange that we POST on development but GET on QA. Solution allows GET requests.

cdoyle-temple (approved):

PR #1948, opened 2021-05-25
#MAN-1095 event recording search returning wrong message

cdoyle-temple:


nomadicoder (changes requested):
There's a stray `binding pry`. Please remove.﻿

nomadicoder (approved):

PR #2076, opened 2021-09-30
#MAN-1130 Enable Accessibility Field for Spaces in Administrate

cdoyle-temple:


dkinzer:
@cdoyle-temple is this the schema.json that we need to update something in catalog too?

cdoyle-temple:
Do you harvest space(s) data from manifold? These are descriptive pages about locations in Charles, such as the One Stop Desk, etc...

PR #2122, opened 2021-11-12
rerun bundle to resolve lock file differences

cdoyle-temple:


sensei100:
@cdoyle-temple Usually the lock file is auto-generated.  Was that change added manually?

cdoyle-temple:
Not sure where the disconnect happened, but the update is also in the two PRs still waiting for review.

cdoyle-temple:
Since this update is in two pending PRs, this PR is unnecessary.

PR #2131, opened 2021-11-17
Menu Category Items fix

cdoyle-temple:
Move login screen out of application layout--this allows us to login even if there is an error in the application layout
Simplifies menu-items exclusion of non-categories
Category name will now defer to label in view (the non-category did not respond to name)
Deprecation warning cleanups for specifying partials


nomadicoder (changes requested):
Are you testing for the conditions that caused the failure we encountered?

cdoyle-temple:
No, just cleaning up so we can get to the longer term solution of separating these menu categories out. That will enable us to create appropriate tests.

cdoyle-temple:
Accidentally pushed up test investigations.

nomadicoder (approved):

PR #2301, opened 2022-03-30
Man 1169 create the ability to modify the order of document attachments to a webpage

cdoyle-temple:


app/views/webpages/home.html.erb
```diff
@@ -1,6 +1,6 @@
 <div id="home">
 <div id="searchFields">
-
+<% binding.pry %>
```
nomadicoder, line 3: Do we need this?

nomadicoder (changes requested):
See my in-line comment

app/views/webpages/home.html.erb
cdoyle-temple: nope

nomadicoder (approved):

PR #2318, opened 2022-04-14
Man 1172 styling change for custom pages and throughout the site

cdoyle-temple:
Adds the alerts


app/assets/javascripts/listeners.js
```diff
@@ -33,7 +33,7 @@ $(document).on('turbolinks:load', function() {
 });
 
 $(document).ready(function(){
-
+  // wpvi
```
sensei100, line 36: @cdoyle-temple I'm not familiar with wpvi.  What does this mean?

sensei100 (commented):
@cdoyle-temple Will any of the header-alert updates affect Library Search?

app/assets/javascripts/listeners.js
cdoyle-temple: This section is for the WPVI videos that we host in the site.

cdoyle-temple:
None of the styling I've done here will affect the header alerts display or functionality in library search.

sensei100:
@cdoyle-temple The alert changes that I was referring to don't seem to be styling changes.  They are updates in the controllers.

app/assets/javascripts/listeners.js
cdoyle-temple: I believe the alerts you look for in the JSON are Alert.where(published: true).where(for_header: false)

Nothing about that has changed.

cdoyle-temple:
@sensei100 I believe the alerts you look for in the JSON are Alert.where(published: true).where(for_header: false)

Nothing about that has changed.

sensei100 (approved):

PR #2710, opened 2023-01-26
MAN-1242 add local search to finding aids page

cdoyle-temple:


cdoyle-temple:
Requested to only duplicate search functionality already within librarysearch. Leaving this branch as an example of how to search action_text fields.

PR #2860, opened 2023-05-25
MAN-1282 Write alerts.json to active record

nomadicoder:
Access alerts json with https://library.temple.edu/alerts_json/show.json


cdoyle-temple (approved):
@nomadicoder This looks good to me now. Nothing changed for the librarysearch ingest and we have the db values available at the /alerts.json endpoint.

PR #2952, opened 2023-08-14
MAN-1300 hours page text change

cdoyle-temple:
Allows for Title, textual Intros, and email recipients to be specified directly from the CMS instead of requiring code changes.


app/views/mail_form/contact.erb
```diff
@@ -1,6 +1,7 @@
 <h4 class="text-decoration-underline"><%= message.subject %></h4>
 <% array_values = "" %>
-<% @resource.mail_form_attributes.each do |attribute, value| next if value.blank? %>
+<% binding.pry %>
```
nomadicoder, line 3: @cdoyle-temple Should this be here?

nomadicoder (changes requested):
Please remove the binding.pry

app/views/mail_form/contact.erb
cdoyle-temple: @nomadicoder Got it. Thanks!

nomadicoder (approved):

PR #3009, opened 2023-09-14
Update README.md

cdoyle-temple:
Add ImageMagick as system requirement

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ manifold manages, orchestrates, and displays data about Temple University Librar
 
 ## System Requirements
 
-- Ruby 2.7.2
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,7 @@ manifold manages, orchestrates, and displays data about Temple University Librar
 
 - Ruby 2.7.2
```
sensei100, line 8: Maybe we should update this version since we are making changes?

sensei100: @cdoyle-temple I think it would make sense to tell users to use whatever Ruby version is in the Gemfile so that you don't have to worry about updating the Readme whenever it changes.

sensei100 (approved):

PR #3020, opened 2023-09-20
Revert "Update access_description_serializable.rb"

cdoyle-temple:
testing nature of error as it can't be reproduced locally


cdoyle-temple:
fine as is

PR #3021, opened 2023-09-20
DEVO-590: docker image smaller

dkinzer:
* Remove unused build arguments.
* Make sure rails env is production to removal of build packages
      happens.
* Removes yarn cache.

This only shaves off some of the load. The rest of the packages are
required so therefore we cannot make it smaller.


sensei100:
@dkinzer I think you based this off the main branch, but this project still has a QA branch.

Gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
-ruby "3.1.4"
+ruby "3.1.4" unless ENV["K8"] == "yes"
```
sensei100, line 5: @dkinzer I'm confused by this part.  We had to update the Ruby version the other day because the docker image was using 3.1.4.  What is it using now?

sensei100 (approved):

PR #3029, opened 2023-09-22
Update qa_deploy image

sensei100:


cdoyle-temple (approved):

dkinzer:
@sensei100 .. actually a quick question about this: we set IMAGE invariable in other projects why aren't we doing that here?

PR #3053, opened 2023-09-29
DEVO-873 Fix skipped tests in some of the models

sensei100:


spec/models/blog_spec.rb
```diff
@@ -4,51 +4,54 @@
...
+            expect(blog.public_status).to be false
+          end
         end
-      end
-      context ":feed_path" do
-        it 'returns "/feed"' do
-          expect(blog.feed_path).to eql "/feed"
+        context ":feed_path" do
+          it 'returns "/feed"' do
+            expect(blog.feed_path).to eql "/feed"
```
cdoyle-temple, line 36: It feels like this should be blog.feed_path rather than the actual value. If the factory changes, then the value will change and the test will fail. That would keep changes in one place only without consequence.

spec/models/blog_spec.rb
```diff
@@ -4,51 +4,54 @@
...
-          # custom method on model not accepting feed_path value from factory and thus sets value to "/feed"
-          expect(FactoryBot.build(:blog_with_feed_path).feed_path).to eql "/weird-feed"
-        end
+    end
+
+    context ":feed_path" do
+      let(:blog) { FactoryBot.build(:blog_with_feed_path) }
+
+      it 'returns "/weird-feed"' do
+        expect(blog.feed_path).to eql "/weird-feed"
```
cdoyle-temple, line 54: Same comment as above,

app/models/building.rb
```diff
@@ -59,6 +59,7 @@ def additional_schema_dot_org_attributes
       email:,
       geo: coordinates,
       googleId: google_id,
+      hours:
```
cdoyle-temple, line 62: Why are we adding hours?

spec/helpers/application_helper_spec.rb
```diff
@@ -162,7 +162,7 @@
 
       it "adds a list item" do
         allow(category).to receive(:items).and_return([items])
-        expect(helper.get_items(category)).to include("<li class=\"mt-0\">")
+        expect(helper.get_items(category)).to include("Charles Samuel Addams Library")
```
cdoyle-temple, line 165: It feels like this should be building.label rather than the actual value. If the factory changes, then the value will change and the test will fail. That would keep changes in one place only without consequence.

spec/models/building_spec.rb
```diff
@@ -88,14 +88,19 @@
     let(:city) { "Philadelphia" }
     let(:state) { "PA" }
     let(:zip_code) { "19122" }
-    let(:building) { FactoryBot.create(:building, address2: "#{city}, #{state}, #{zip_code}") }
+    let(:building) { FactoryBot.create(:building, address2: "#{city}, #{state}, #{zip_code}", description: "building description", hours: "paley") }
```
cdoyle-temple, line 91: Hours are not part of the serialization.

app/models/building.rb
sensei100: @cdoyle-temple Do you want that test removed then?

spec/helpers/application_helper_spec.rb
sensei100: I updated this one.

spec/models/blog_spec.rb
sensei100: I changed the building.label test, but I think the blog tests should stay the way they are.  We are testing 2 different contexts and expecting different text to be returned for each one.  I think using the expected text makes it easier to understand what the expectations are.

sensei100:
@cdoyle-temple Can you review this again?

app/models/building.rb
cdoyle-temple: Yes, it should not have been included.

cdoyle-temple (approved):

PR #3071, opened 2023-10-06
DEVO-882 Cover missing test coverage in Finding Aids

nomadicoder:


cdoyle-temple (changes requested):
@nomadicoder This test reduces the amount of coverage on the finding aids helper from 88.89% to 25%. This reduction can be seen in coveralls for app/helpers/finding_aids_helper.rb.

cdoyle-temple (approved):

PR #3096, opened 2023-10-26
MAN-1281: Fix error when destroying external links.

dkinzer:


app/views/layouts/devise/sessions.html.erb
```diff
@@ -4,7 +4,7 @@
       <meta http-equiv="refresh" content="0; url=<%= account_google_oauth2_omniauth_authorize_url %>" />
   <% end %>
 <% end %>
-<% if Rails.env != "development" %>
+<% if Rails.env == "production" %>
```
cdoyle-temple, line 7: @dkinzer We don't want these fields in production, only on local environments.

cdoyle-temple (changes requested):

cdoyle-temple:
@dkinzer External links which are currently attached to another model (e.g. a Webpage) are now not being destroyed when the button is clicked. There's also no error message or other indication that I tried to destroy the link.

dkinzer:
@cdoyle-temple I addressed the issue and added a test to check that when a link is attached to another model the link will not get destroyed.

dkinzer:
> @dkinzer External links which are currently attached to another model (e.g. a Webpage) are now not being destroyed when the button is clicked. There's also no error message or other indication that I tried to destroy the link.

Just a note that in fact for the test you ran the link was not being destroyed.  But, that is the expected behavior for when a link is attached to a model.  The issue was that by removing the redirect, the notice was not appearing.  The fix was to made sure that the redirect happened inside the block that checks for external links in other models and not to happen at the end after `super` is  called that has it's own redirect.

cdoyle-temple (approved):

PR #3101, opened 2023-10-27
DEVO-873 Updates pending specs for some  models

sensei100:
- Adds tests for form and form_info models
- Removes pending line from exhibitions because it does have tests
- Removes unused methods from group model
- Refactors validation tests for groups

### DIFF
--- a/spec/factories/groups.rb
+++ b/spec/factories/groups.rb
@@ -15,6 +15,7 @@ FactoryBot.define do
         webpage.file_uploads << file
       end
     end
+
     trait :with_files do
       after :create do |webpage|
         file = FactoryBot.create(:file_upload)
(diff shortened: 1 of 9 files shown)


spec/models/form_info_spec.rb
```diff
@@ -3,5 +3,18 @@
...
+
+    it "is valid with valid attribues" do
+      expect(form_info).to be_valid
+    end
+  end
+
+  describe "invalid form_attributes" do
+    let(:form_info) { FactoryBot.build(:form_info) }
+    it "is not valid without a recipient" do
+      expect(form_info).to be_valid
```
cdoyle-temple, line 17: The expect is looking for validity, but the description says it should be invalid.

spec/factories/groups.rb
```diff
@@ -23,5 +24,13 @@
...
       end
     end
+
+    trait :no_chair_dept_heads do
+      chair_dept_heads { [] }
+    end
+
+    trait :no_space do
+      space {}
+    end
```
cdoyle-temple, line 34: :no_space trait isn't used in any of the tests, is it necessary?

cdoyle-temple (changes requested):

sensei100:
@cdoyle-temple Fixed both issues that you found.  You have good eyes!

cdoyle-temple (approved):

PR #3108, opened 2023-10-30
MAN-1313: Fix groups not saved when adding spaces

dkinzer:
- [ ] Depends on running `rails data_migration:migrate_space_id` to move `space_id` from `SpaceGroup`  to `Group`.


cdoyle-temple (changes requested):
@dkinzer Can you retain the order by name in the select list?

spec/features/admin/admin_groups_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+
+      visit admin_groups_path + "/#{ group.friendly_id }"
+      expect(page).to have_link(group.space.name)
+
+      visit admin_spaces_path
+      expect(page).to have_link(group.space.name)
+
+      click_link "Destroy"
+      expect(page).not_to have_link(group.space.name)
```
cdoyle-temple, line 79: @dkinzer Shouldn't this throw an error that the space is still attached to a model? Or is that something for another ticket?

dkinzer: @cdoyle-temple that's not the current behavior on qa which is why I specifically added the `dependent: :nullify` option to the `has_many groups` directive in the Stage model.  That said if you want it error out we can do that too, but note that currently qa and I assume live simply do not error out when you delete a space that have groups referencing it.

dkinzer:
@cdoyle-temple sorted the the spaces options by name ASC.  Also I added a spec that checks for that so that it's not lost in the future.

cdoyle-temple (approved):

PR #3111, opened 2023-10-31
Use main branch for qa deployments.

dkinzer:


sensei100 (commented):
@dkinzer This would run everything using the stage inventory in the playbook.  That means the hosts would be different.  I think we need to update the playbook branch.

dkinzer:
sorry copypated form stage meant qa... was going to close this anyway

PR #3134, opened 2023-11-13
MAN-1353 add snippet feature to edit top text on etexts page

cdoyle-temple:


ebtoner:
@cdoyle-temple - can you clarify what "past-event-videos-intro" is in this context?

cdoyle-temple:
@ebtoner That's the identifier for the snippet which provides the text and the title of the page. Snippets are a model we use so backend users can update text they'd normally have to ask for a code change for. The snippet and whichever polymorphic model, in this case a view component, are associated through this identifier (a friendly id rather than a numeric id), and as long as this friendly id remains the same, the display text can be changed at will without a code change.

ebtoner:
@cdoyle-temple -- ok, that's fine. I was just the labelling itself that threw me off. If this is used across features/page types, maybe we should think about changing the id label to something more generic? I'll go ahead and approve this change, though.

ebtoner (approved):

PR #3136, opened 2023-11-14
Update MASTER key to use MANIFOLD key

cdoyle-temple:


.docker/app/Dockerfile
```diff
@@ -6,13 +6,13 @@ COPY . .
 
 USER root
 
-ARG RAILS_MASTER_KEY \
+ARG RAILS_MASTER_KEY=MANIFOLD_MASTER_KEY \
```
sensei100, line 9: I think we need to include a $ in front of MANIFOLD_MASTER_KEY

cdoyle-temple: done

sensei100 (approved):

PR #3137, opened 2023-11-14
disable skylight for centos

cdoyle-temple:


.docker/app/Dockerfile
```diff
@@ -26,7 +26,6 @@ RUN apk add -U --no-cache \
       nodejs=18.18.2-r0 \
       yarn=1.22.19-r0 && \
     if [ "${RAILS_ENV}" = "production" ]; then bundle config set --local without "development test"; fi && \
-    bundle update --conservative skylight && \
```
dkinzer, line 29: This is just for docker it does not affect production

PR #3366, opened 2024-05-01
Finding aids json location

cdoyle-temple:


sensei100 (approved):

config/routes.rb
```diff
@@ -116,7 +116,7 @@
     get "finding_aids" => :index
     get "finding_aids/:id" => :show
     get "finding-aids/:id" => :show
-    get "finding_aids.json", to: redirect("assets/finding_aids.json")
```
dkinzer, line 119: @sensei100 @cdoyle-temple just an fyi that we didn't make the assets folder writable just /app/public/cache, /tmp, /app/log and app/tmp

sensei100: @dkinzer I think it should be okay because he removed that line.

PR #3382, opened 2024-05-08
update libjpeg version

cdoyle-temple:


sensei100:
@cdoyle-temple Can you also try adding gcompat=1.1.0-r4?  That package is in the tupress Dockerfile.

sensei100 (approved):

PR #3466, opened 2024-07-10
DEVO-1099: Add a redirect from link_exchange to tulle.tul-infra.page

dkinzer:


config/routes.rb
```diff
@@ -4,6 +4,9 @@
   mount Rswag::Api::Engine => "/"
   mount Rswag::Ui::Engine  => "api-docs"
 
+  get "link_exchange/*path", to: redirect { |params, request| "https://tulle.tul-infra.page/#{params[:path]}" }
```
ebtoner, line 7: @dkinzer Can you also add the following?

`get "r/*path", to: redirect { |params, request| "https://tulle.tul-infra.page/r/#{params[:path]}" }`

These are redirect links that are generated by the Link Exchanger Service.

ebtoner: Actually, forget that. Let's first see if it works without this change.

ebtoner (approved):

ebtoner:
Approved, though there is a rubocop error that needs fixing.

PR #3510, opened 2024-08-09
BL-1410 Remove resource_params method override in admin application_controller

nomadicoder:
- resource_params already provided by administrate's application controller which does not have problems with missing primary_key
- Create spec test for this issue


nomadicoder:
Wrong Git Branch

PR #3522, opened 2024-08-14
MAN-1410 - Remove main finding_aids resource from routes

nomadicoder:
- Comments out the main finding_aids routes
- Disable specs deending on legacy finding_aids resources
- If this works, need to remove finding_aids from application


nomadicoder:
We will maintain existing finding_aids on Manifold.  We only want to redirect the finding_aids index page to the new site.

PR #3543, opened 2024-08-27
MAN-1408 Add ability to enable and disable forms

nomadicoder:


nomadicoder:
@cdoyle-temple Is this what we want here?  It's draft because I'd like to see if I can move the predicate out of the view into the controller.  Also want to clean up the ugly nested `if` block.

PR #3733, opened 2025-01-14
Update Event Space Request form

nomadicoder:
- Emphasize Event Start Time and Department/Organization
- Change 3rd-floor event space text
- Change Catering text (initial work)


cdoyle-temple (changes requested):
Needs more updates. Will confer offline via Slack

cdoyle-temple (approved):

PR #3768, opened 2025-02-11
MAN-1459 add digital collections link

nomadicoder:
Awaiting update to digital collections icon.


nomadicoder:
Broken branch.  Will create new one.

PR #3769, opened 2025-02-11
MAN-1459 add digital collecitons link to main page

nomadicoder:


nomadicoder:
There is a link to digital collections in the config/locale/en.yml.  I should replace the hard link in app/views/webpages/home.html.erb with that.

cdoyle-temple (approved):

PR #3788, opened 2025-02-18
MAN-1422 add limit to events button

nomadicoder:
Filter on non-workshops, which is returned when clicking on the "Limit to Events" button.


app/views/events/index.html.erb
```diff
@@ -51,6 +51,9 @@
       <div class="col-12">
         <%= workshops_link(params[:type])%>
       </div>
+      <div class="col-12">
```
cdoyle-temple, line 54: These should appear horizontally, with a pipe separator. Responsive views can wrap links if necessary.

config/locales/en.yml
```diff
@@ -480,6 +480,8 @@ en:
         upcoming_events: Upcoming Events
         events: Events
         events_on: Events on
+        workshops: Workshops
+        not_workshops: Events
```
cdoyle-temple, line 484: Can these be set to Workshop Events, and Non-Workshop Events?

cdoyle-temple (approved):

PR #3833, opened 2025-03-14
Todays hours cached

cdoyle-temple:


cdoyle-temple:
This work will need a cron job setup in K8s before deploying

dkinzer (approved):

PR #3891, opened 2025-05-01
MAN-1477 library temple accessibility review

cdoyle-temple:
Updates from dubbot scans. Most actionable issues are addressed. Will rerun scan after next release to evaluate.


app/views/webpages/_tutorial.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="mt-4 temple-library-tutorial">
-  <iframe width="850px" height="650px" src="https://alt.library.temple.edu/tutorials/<%= @webpage.tutorial_path %>"></iframe>
+  <iframe width="850px" height="650px" aria-label="<%= @webpage.label %> tutorial" src="https://alt.library.temple.edu/tutorials/<%= @webpage.tutorial_path %>"></iframe>
```
sensei100, line 2: @cdoyle-temple Is this missing a beginning quote?

app/views/webpages/_virtual-tour.html.erb
```diff
@@ -1,5 +1,5 @@
 <div id="virtual-tour">
-  <iframe src="<%= @webpage.virtual_tour %>" name="TOUR NAME" 
+  <iframe src="<%= @webpage.virtual_tour %>" name="TOUR NAME" aria-label="Virtual Tour
```
sensei100, line 2: @cdoyle-temple Is this missing an end quote?

sensei100:
@cdoyle-temple For the alt-texts="" is there a way to add something there?  That doesn't seem like it adds anything to help with accessibility.  I wondered the same thing for the alt="pdf" - does that document change?  If not, could the alt text be more specific?

On a broader note, should we add a form field for alt text for people to complete when they add images to the website?

app/views/webpages/_tutorial.html.erb
cdoyle-temple: no, it looks well-formed

app/views/webpages/_virtual-tour.html.erb
cdoyle-temple: it is, thanks!

cdoyle-temple:
@sensei100 When the alt text is set to "" screen readers know to skip over the image as decorative. This is a standard way of handling decorative images so that screen readers aren't repeating information unnecessarily and the image tag is not read out to the user with no context. When we have images that relay information that is not in the accompanying text, via caption, or context within a page, we should have an alt text.

cdoyle-temple:
We do not have a way to add alt text to images inserted by users into our webpages. Our editor does not have that functionality. But it does allow captions, which are acceptable alternatives to alt text. We had one instance of that flagged in the scan but it did have the caption filled out. https://library.temple.edu/services/ginsburg-poster-printing-service

sensei100:
@cdoyle-temple The reason I was asking about whether the file would change is because "pdf file" doesn't give the user any context about what the file is.  If the file doesn't change, I think the alt text should explain the purpose or context of the file rather than the file type.

cdoyle-temple:
The file type doesn't change, but the files do. Those files are attachments. There can be multiple attachments per page: https://library.temple.edu/webpages/jobs. 

The icon without the alt text lets the user know the link is to a PDF, a visual indicator that should have a text equivalent. So I added the alt tag. PDFs can be an issue for accessibility so it's nice to be forewarned. So can Word docs but luckily we don't have any of those!

sensei100 (approved):

PR #3957, opened 2025-07-09
add tmp folder addition to dockerfile

cdoyle-temple:


nomadicoder (approved):

.docker/app/Dockerfile
```diff
@@ -43,11 +43,11 @@ RUN apk add -U --no-cache \
...
     rm -rf /var/cache/apk/ && \
-    rm -rf /usr/local/share/.cache/yarn && \
-    chown nobody -R /app/tmp && \
-    chown nobody -R /app/log
+    rm -rf /usr/local/share/.cache/yarn  && \
+    mkdir -p /secure-tmp && chmod 700 /secure-tmp && \
+    mkdir -p /secure-tmp/log && chmod 700 /secure-tmp/log
 
-USER nobody
+USER 65534
```
dkinzer, line 50: I would prefer if user nobody is kept.

.docker/app/Dockerfile
```diff
@@ -43,11 +43,11 @@ RUN apk add -U --no-cache \
     if [ "${RAILS_ENV}" = "production" ]; then rm -rf tmp/*; fi && \
     if [ "${RAILS_ENV}" = "production" ]; then apk del build-dependencies; fi && \
     rm -rf /var/cache/apk/ && \
-    rm -rf /usr/local/share/.cache/yarn && \
-    chown nobody -R /app/tmp && \
-    chown nobody -R /app/log
+    rm -rf /usr/local/share/.cache/yarn  && \
+    mkdir -p /secure-tmp && chmod 700 /secure-tmp && \
```
dkinzer, line 47: I think this should be added to entry-point script because volume mounts are not availabe until after the image is being used in a container.

.docker/app/Dockerfile
sensei100: @dkinzer The best practice recommendation is to use the number because  nobody may resolve to different UIDs depending on the OS base image. We are also using fsGroups in our securityContexts new and nobody isn't allowed there, you have to use the number.  So I thought it felt better to keep things consistent and stick with one user.

PR #4014, opened 2025-08-15
Revert "DB Seed per deploy"

sensei100:
Reverts tulibraries/manifold#4013


sensei100:
I think there are some potential risks in putting this command directly in a Dockerfile, so I would rather find another way to do this.

PR #4060, opened 2025-09-29
MAN-1498-space-hours-widget-not-functioning

cdoyle-temple:


sensei100 (approved):

sensei100:
> @sensei100 A test that downcase works?

Yes, since people were having a problem with this working I think it makes sense to add a test to ensure that this case works as expected now.

PR #4315, opened 2026-04-07
MAN-1533 punctuation rendering in event titles

cdoyle-temple:


sensei100:
@cdoyle-temple I think this still needs the CGI.unescapeHTML in order to handle this: &nbsp;

cdoyle-temple:
@sensei100 Nokogiri does this already when its .text method is called, so it would be redundant to use both there, but in cases where it can't parse the xml and we rely on the fetch to get the value, an escaped character could slip through. I added a catch for that using the CGI.unescapeHTML.

sensei100:
@cdoyle-temple Sorry, it looks like I didn't paste what I meant to in my original comment.  I'm talking about making sure this gets taken care of: `&nbsp;`.  Can you add a test for that case?

cdoyle-temple:
@sensei100 Yes, a test case was added: spec/services/sync_events_spec.rb:51

cdoyle-temple:
@sensei100 `On further research, Ruby's CGI.unescapeHTML only decodes 5 HTML entities, not including &nbsp;. Nokogiri::HTML4.fragment covers the full HTML entity library and all Unicode characters. I substantially simplified the workflow and have tests in place for HTML-entity-encoded characters and Unicode non-breaking spaces — but the fix covers all HTML entities and Unicode, not just those two cases. Also covers percent-encoded values via CGI.unescape.`

sensei100 (approved):

PR #4338, opened 2026-04-17
DEVO-1362 upgrade manifold

cdoyle-temple:


app/javascript/controllers/index.js
```diff
@@ -1,6 +1,28 @@
...
+import FormSubmissionsController from "controllers/form_submissions_controller"
+import JumbleController from "controllers/jumble_controller"
+import PeopleController from "controllers/people_controller"
+import PrintController from "controllers/print_controller"
+import SearchFormController from "controllers/search_form_controller"
+import SelectizeController from "controllers/selectize_controller"
+import WpviController from "controllers/wpvi_controller"
 
-import controllers from "./**/*_controller.js"
-controllers.forEach((controller) => {
```
sensei100, line 4: @cdoyle-temple I think it is better to use the forEach instead of having to add a new registration every time a controller is created.  I think devs will forget about that step.  Can we try to make that work?

app/javascript/homepage.js
```diff
@@ -1,20 +1,20 @@
...
+import FailoverMobileController from "controllers/failover_mobile_controller"
+import HighlightsCarouselController from "controllers/highlights_carousel_controller"
+import JumbleController from "controllers/jumble_controller"
+import NewsCarouselController from "controllers/news_carousel_controller"
 
-application.register("jumble", JumbleController)
 application.register("external-links", ExternalLinksController)
-application.register("news-carousel", NewsCarouselController)
-application.register("highlights-carousel", HighlightsCarouselController)
-application.register("digcols-carousel", DigcolsCarouselController)
```
sensei100, line 16: Could we get the loop working here too?

sensei100 (changes requested):

app/javascript/homepage.js
cdoyle-temple: Yes, I was able to divide the locations of the controllers to effect the globular loading (via eager-loading of controllers at the entry-points) of each, similar to what was in place previously but now with the split between homepage and application (with consideration for controllers both might need).

sensei100 (approved):

PR #4413, opened 2026-06-11
Cloudflare Turnstile Forms Integration

cdoyle-temple:


sensei100:
@cdoyle-temple Can you please put this behind Flipflop so that if there are issues we can quickly turn it off?

sensei100 (changes requested):
See comment.

dkinzer:
@cdoyle-temple  it seems  you are not using bot_challenge_page gem but are instead rolling out your own implementation is that correct?

sensei100:
@cdoyle-temple The bot_challenge gem was specifically written for Rails applications by another institution.  I do want to use that here as well.

cdoyle-temple:
@dkinzer Yes, the scope for this was just the public facing forms, one endpoint. It seemed simpler to implement this way due to the difference between app needs. But I can switch it over.

dkinzer:
@cdoyle-temple you can configure scope with bot_challenge_page gem by setting :only attribute in the mixin.

example `bot_challenge only: :show`   or `bot_challenge only: show, if: ->  { Flipflop.bot_challenge? }`

cdoyle-temple:
@sensei100 @dkinzer bot_challenge_page only covers GET routes, and our form submissions happen on POST. So I'd like to keep the existing Turnstile verification on the form POST to protect the submission itself, but still allow the form show page to be protected by the GET-side bot challenge.

cdoyle-temple:
> @cdoyle-temple you can configure scope with bot_challenge_page gem by setting :only attribute in the mixin.
> 
> example `bot_challenge only: :show` or `bot_challenge only: show, if: -> { Flipflop.bot_challenge? }`

I can add this for the show page, but we also need to protect the form submissions on create, POST requests (bot_challenge_page only works for GET routes). We also don't have flipflop on manifold, but I've configured the service to be enabled only when the env var from vault, TURNSTILE_ENABLED is true and there are valid secrets there as well.

dkinzer:
OK, I didn't realize bot_challenge_page gem does not work on POST. The developer of that gem didn't need that functionality for his use case, it turns out. I'm fine just keeping your solution instead of having two different solutions for both use cases.

cdoyle-temple:
We could do both, but that would mean challenging users twice. As is, Turnstile runs on the form show page, Cloudflare adds its response token to a hidden field in the form, and that token is submitted with the POST. On create (the submission action), the server verifies the token with Cloudflare before the submission is accepted. So the verification is already happening as part of the actual form submission flow.

cdoyle-temple:
@sensei100 I need to know if this approach is ok, and if we still need flip-flop for this one feature if we can control via env vars from vault.

sensei100:
@cdoyle-temple I'm okay with this approach since the bot_challenge gem doesn't support post requests.  But I do still want it behind Flipflop.

sensei100 (approved):

PR #4418, opened 2026-06-12
Cloudflare Turnstile Forms Integration

cdoyle-temple:
Move from cookie to activerecord so feature applies to all users.

### DIFF
--- a/config/features.rb
+++ b/config/features.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Flipflop.configure do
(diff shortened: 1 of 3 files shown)


dkinzer:
@cdoyle-temple you can add all three strategies like we do in tul_cob then who ever enables this feature of future feature can pick how to enable it.

cdoyle-temple:
@dkinzer Ok

dkinzer (approved):

